PR #3, opened 2020-11-11
refactor: basic conventions

nugaon:
## Features

- added eslint rules, prettier and tsconfig to the project,
- refactor the already existing code by that,
- created `npm run format` command for rewrite all ts files in src/** and test/**
- defined VSCode settings for auto-correcting on save and use Prettier 

## TODO
- [x] Redesign GitHub Workflow (browser tests)
- [x] Change README
- [x] Webpack integration
- [x] Replace Karma with Jest
- [x] Remove Tasegir dependency
- [x] Browser tests


agazso:
I think it would be better to merge this to the `main` branch now

package.json
```diff
@@ -12,59 +12,71 @@
...
   },
   "repository": {
     "type": "git",
     "url": "https://github.com/ethersphere/bee-js.git"
   },
   "scripts": {
     "postinstall": "patch-package",
-    "compile": "tasegir compile",
+    "compile": "npm run compile:node",
     "compile:watch": "tasegir compile -- --watch",
```
vojtechsimetka, line 26: remove

package.json
```diff
@@ -12,59 +12,71 @@
...
+    "compile:node": "webpack --progress --mode development --env.target=node",
+    "compile:browser": "webpack --progress --mode development --env.target=web",
     "docs": "typedoc --mode modules --excludeNotExported --readme none --out docs src",
-    "types-check": "tasegir types-check",
-    "exec": "tasegir run",
-    "lint": "tasegir lint",
+    "types-check": "tsc --noEmit",
     "release": "tasegir release",
```
vojtechsimetka, line 31: remove

package.json
```diff
@@ -12,59 +12,71 @@
...
-    "./lib/utils/data.js": "./lib/utils/data.browser.js",
-    "./src/utils/data.ts": "./src/utils/data.browser.ts"
+    "stream": "readable-stream"
   },
   "repository": {
     "type": "git",
     "url": "https://github.com/ethersphere/bee-js.git"
   },
   "scripts": {
     "postinstall": "patch-package",
```
vojtechsimetka, line 24: remove

src/utils/data.ts
```diff
@@ -1,11 +1,22 @@
...
+    const buffers: Array<Uint8Array> = []
+    data.on('data', d => {
+      buffers.push(d)
+    })
+    data.on('end', () => {
+      resolve(Buffer.concat(buffers))
+    })
+  })
+
+  return readableData
```
vojtechsimetka, line 21: This means, that from now on the nodejs implementation also does not utilise data streaming, and instead puts everything in buffer, right?

.mocharc.yaml
```diff
@@ -0,0 +1,13 @@
+# This is an example Mocha config containing every Mocha option plus others
```
vojtechsimetka, line 1: Do we need mocha config when we are using Jest?

karma.conf.ts
```diff
@@ -0,0 +1,169 @@
+
```
vojtechsimetka, line 1: Do we need karma config when we are using Jest?

package.json
```diff
@@ -12,59 +12,71 @@
...
-    "exec": "tasegir run",
-    "lint": "tasegir lint",
+    "types-check": "tsc --noEmit",
     "release": "tasegir release",
-    "build": "tasegir build",
-    "test": "tasegir test",
-    "test:node": "tasegir test --target node",
-    "test:browser": "tasegir test --target browser --target webworker",
-    "tasegir:link": "ln -s ./node_modules/tasegir/src/config/tsconfig.json ./tsconfig.json && ln -s ./node_modules/tasegir/src/config/eslintrc.js ./.eslintrc.js"
+    "test": "jest --config=jest.config.ts",
```
vojtechsimetka, line 32: Could we separate browser and nodejs tests?

vojtechsimetka (changes requested):

package.json
agazso: I propose let's merge this and open issues for the topics that @vojtechsimetka raised.

agazso:
I propose let's merge this and open issues for the topics that @vojtechsimetka raised.

agazso (approved):
I propose let's merge this and open issues for the topics that @vojtechsimetka raised.

vojtechsimetka:
I agree

vojtechsimetka (approved):

package.json
nugaon: The building process could read the libraries/files that differ on NodeJS<->Browser side from this list. Although it is not implemented yet.

package.json
nugaon: Yes, we can

karma.conf.ts
nugaon: No, we don't need I will delete it in the next commit

.mocharc.yaml
nugaon: No, we don't need I will delete it in the next commit

package.json
nugaon: Yes, it is possible, I put that into the NPM scripts in the next change

PR #9, opened 2020-11-22
Collection support

AuHau:
This is a bit rough draft PR that adds support for Collection to your modules and does few other small fixes and tweaks as well like:

 - fixing `package.json` and its script for using it as npm's Git dependency
 - fixing upload limitation of file's size of axios 
 - adds Collection support that allows streaming the upload to the node 

What is missing atm. is mainly browser support as streaming is not available in browser atm. So most probably buffering should be employed.

Feel free to pick on things you want to have and reject the other :-)


agazso:
Added collection functionality based on this PR in #28

PR #14, opened 2020-11-26
refactor: test utility function randomBuffer is now pseudorandom

### LINKED ISSUE #13: Remove randomness from tests
See https://github.com/ethersphere/bee-js/pull/1#discussion_r520036713

vojtechsimetka:
Resolves #13

### DIFF
--- a/test/utils.ts
+++ b/test/utils.ts
@@ -21,17 +21,28 @@ export function createReadable(input: string): Readable {
 }
 
 /**
+ * Lehmer random number generator with seed (minstd_rand in C++11)
+ * !!! Very fast but not well distributed pseudo-random function!!!
+ *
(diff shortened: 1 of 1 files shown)


test/utils.ts
```diff
@@ -28,10 +38,11 @@ export function createReadable (input: string): Readable {
  * @param length
  */
 export function randomBuffer (length: number): Buffer {
+  const rand = lrng(500)
```
agazso, line 41: One potential issue with this approach is that if you want to use `randomBuffer` more than once  in a test (for example to generate two random files) it will always return the same content. I think it would be better if the seed could passed in as a second argument with a default value.

agazso (approved):

PR #17, opened 2020-12-03
refactor: basic conventions

nugaon:
## Features

- added eslint rules, prettier and tsconfig to the project,
- refactor the already existing code by that,
- created `npm run format` command for rewrite all ts files in src/** and test/**
- defined VSCode settings for auto-correcting on save and use Prettier 

## TODO
- [x] Redesign GitHub Workflow (browser tests)
- [x] Change README
- [x] Webpack integration
- [x] Replace Karma with Jest
- [x] Remove Tasegir dependency
- [x] Browser tests
- [ ] make Jest alias same as in webpack (browser<->node)


.gitignore
```diff
@@ -1,7 +1,4 @@
 lib
-types
```
vojtechsimetka, line 2: Would we now be packing types directly and not generating them?

vojtechsimetka (approved):

.gitignore
nugaon: sometimes it's necessary, for example when u'll write ur own Jest matchers.

PR #28, opened 2020-12-08
feat: module collection

agazso:
This PR introduces the collection functionality, e.g. uploading multiple files and then accessing files from the collection.

I plan to not fix the issues with the `Readable` interface because I discovered that we may need to choose a different approach for the browser and it will most likely affect that as well. The issues with the browser are the following:
- we are using `tar-stream` for creating the tar in a streaming way
- it relies on dependencies that are not working in the browser
  - for example it uses `buffer` which is not polyfilled from version 5 of webpack (that is the version we use)
- it relies also on `readable-stream` which relies on `stream` which is also not polyfilled
- and the list goes on

I also found that the generated JS bundle for the browser is already quite big.

Based on our previous conversations I would suggest a different approach, to try to use browser APIs and polyfill them on Node.js. On the server side the size doesn't matter that much anyhow. For example using something like this:
https://github.com/MattiasBuelens/web-streams-polyfill

This would require more research and maybe changing some packages that we depend on now but I would keep it as a separate issue and not as part of this PR.


src/types.ts
```diff
@@ -22,3 +31,35 @@ export interface Tag {
...
+
+  /**
+   *
+   */
+  path: string
+
+  /**
+   * Required when `data` is Readable.
+   */
+  size?: number
```
nugaon, line 59: it could be better to create type, when the data is Readable and the size is required.

src/utils/headers.ts
```diff
@@ -0,0 +1,52 @@
+import contentDisposition from 'content-disposition'
+
+import { Dictionary, FileHeaders, UploadHeaders, UploadOptions } from '../types'
+import { BeeError } from './error'
+
+function readContentDispositionFilename(header?: string): string {
+  try {
+    if (header == null) {
+      throw new BeeError('missing content-disposition header')
```
nugaon, line 9: why is the header parameter optional then? it means it can be undefined, but you don't handle it, just for null value.

src/utils/headers.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+    throw new BeeError('invalid content-disposition header')
+  } catch (e) {
+    throw new BeeError(e.message)
+  }
+}
+
+function readTagUid(header?: string): number | undefined {
+  if (header == null) {
+    return undefined
```
nugaon, line 24: why is the header parameter optional then? it means it can be undefined, but you don't handle it, just for null value.

src/utils/readable.ts
```diff
@@ -0,0 +1,11 @@
+import { Readable } from 'stream'
```
nugaon, line 1: we should handle the browser side as well. if it's just a type, import with `import type (...)`

src/utils/tar.ts
```diff
@@ -0,0 +1,77 @@
+import { Readable, Writable } from 'stream'
```
nugaon, line 1: we should handle browser-side processing as well with `readable-stream`

nugaon (dismissed):

src/utils/headers.ts
agazso: `== null` checks for both `null` and `undefined` (somewhat surprisingly), so out of habit I usually just use that. I thought this is a common idiom but we can decide to use `== undefined` if you think that's more clear.

src/modules/collection.ts
```diff
@@ -0,0 +1,182 @@
...
+    if (entry.data instanceof Uint8Array) {
+      tar.addUint8Array(entry.path, entry.data)
+    } else if (isReadable(entry.data)) {
+      if (entry.size === undefined) {
+        throw new BeeError(`Size has to be specified for Readable data! [${entry.path}]`)
+      }
+
+      tar.addStream(entry.path, entry.size, entry.data)
+    }
+  }
```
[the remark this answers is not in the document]
agazso: I wouldn't worry about performance here, because generally creating the tar and uploading it is orders of magnitude longer time than the check itself. Besides the code is now very straightforward and I don't really see how a change what you proposed could be implemented without casts.

src/modules/collection.ts
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * Download single file as a buffer from Collection given using the path
+ *
+ * @param url  Bee URL
+ * @param hash Bee Collection hash
+ * @param path Path of the requested file in the Collection
+ */
+export async function download(url: string, hash: string, path = ''): Promise<File<Uint8Array>> {
+  if (!url || url === '') {
+    throw new BeeError('url parameter is required and cannot be empty')
```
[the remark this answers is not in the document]
agazso: Good point

src/modules/collection.ts
```diff
@@ -0,0 +1,182 @@
...
+    entry =>
+      typeof entry !== 'object' || !entry.data || !entry.path || !(isUint8Array(entry.data) || isReadable(entry.data)),
+  )
+}
+
+function packData(data: Collection<Uint8Array | Readable>): TarArchive {
+  const tar = new TarArchive()
+
+  for (const entry of data) {
+    if (entry.data instanceof Uint8Array) {
```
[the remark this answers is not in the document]
agazso: Good point, I will change it

src/utils/headers.ts
nugaon: in this context that would be definitely clearer, but better to use `if (!header)` instead of that

src/modules/collection.ts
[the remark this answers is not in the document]
nugaon: I think the casting is okay if we can save plenty of unnecessary checkings in loops. You're right about this snippet in the whole process is not the most resource consumable task, but we should strive for the optimal running time at every function regardless the whole context.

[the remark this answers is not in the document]
agazso: I don't necessary agree to work on microoptimizations like this without measuring actual performance and seeing that it causes issues. Optimizations in general lead to less maintainable code.

src/types.ts
agazso: See my comment for the PR

src/utils/tar.ts
agazso: See my comment for the PR

src/modules/collection.ts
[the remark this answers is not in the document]
nugaon: In my opinion, this is not micro-optimization and also can be handled in a nice way.

src/modules/collection.ts
```diff
@@ -0,0 +1,200 @@
...
+/**
+ * Upload collection of files to a Bee node
+ *
+ * @param url     Bee URL
+ * @param data    Data in Collection format to be uploaded
+ * @param options Additional options like tag, encryption, pinning
+ */
+export async function upload(
+  url: string,
+  data: Collection<Uint8Array | Readable>,
```
vojtechsimetka, line 123: For browsers I would love to have support for FormData and use Multipart upload

src/modules/collection.ts
```diff
@@ -0,0 +1,200 @@
...
+  return collection
+}
+
+export async function buildFileListCollection(fileList: FileList): Promise<Collection<Uint8Array>> {
+  const collection: Collection<Uint8Array> = []
+
+  for (let i = 0; i < fileList.length; i++) {
+    const file = fileList.item(i)
+
+    if (file != null) {
```
vojtechsimetka, line 103:
```suggestion
    if (file) {
```

I suggest we introduce eslint rule `no-eq-null` https://eslint.org/docs/rules/no-eq-null

vojtechsimetka (approved):
LGTM

src/modules/collection.ts
agazso: Good idea

src/modules/collection.ts
agazso: I'm curious in which case you would prefer it over the FileList API (that we support).

https://developer.mozilla.org/en-US/docs/Web/API/FileList

mshearer0: > NOTE: For browsers I would love to have support for FormData and use Multipart upload. Not within a scope for this PR

Hi, does Bee now support this? I'm struggling with missing Content-type errors

PR #35, opened 2020-12-11
feat: tar upload in the browser

### LINKED ISSUE #32: Remove tar-stream and use in-memory tar for better browser support

agazso:
Changes proposed in this PR:
- Removed the `tar-stream` library dependency in order to be able to build the library for the browser. Instead it uses the `tar-js` library which creates the tar file in memory. With this change we are no longer able to upload huge files but our short term goal is to be able to upload websites and for that this is good enough. For a longer term solution we are evaluating several different approaches and one of them is fixing tar streaming in the browser.
- Also removed the content-disposition library dependency. With only these two changes we reduced the bundle size from ~300K to less than ~100K in the browser.
- Added workarounds for Safari's lack of `ArrayBuffer` support on the `File` interface and also for the nonstandard `webkitRelativePath` properties.
- Renamed our `File` interface to `FileData` because there was a name collision with the browser `File` interface.

Fixes #32


src/modules/collection.ts
```diff
@@ -94,16 +74,41 @@ export async function buildCollection(dir: string, recursive = true): Promise<Co
...
-    const file = fileList.item(i)
+    const file = fileList[i] as WebkitFile
 
-    if (file != null) {
+    if (file) {
       collection.push({
-        path: file.name,
-        data: new Uint8Array(await file.arrayBuffer()),
+        path: filePath(file),
+        data: new Uint8Array(await fileArrayBuffer(file)),
```
vojtechsimetka, line 111: Similarly, this could be already on the interface `WebkitFile` and already return `Uint8Array`

vojtechsimetka (approved):
I know we agreed that for now we will not use streaming and use in memory tar. However, I want to point out that each file is stored in memory twice:
- once when read from the filesystem [collection:57](https://github.com/ethersphere/bee-js/pull/35/files#diff-e49f543a389f6fde536d9872dbd29d7373fd6186219fa95dacf879817b0309ddR67)
- second time when creating in memory tar [tar:6-12](https://github.com/ethersphere/bee-js/pull/35/files#diff-e24585393060b7edf90d6527e7c9b16e681c030919a4b46de4a175ee36b8a367R6-R11)

I wonder if we need to buffer the file already in the collections.

agazso:
> I know we agreed that for now we will not use streaming and use in memory tar. However, I want to point out that each file is stored in memory twice:
> 
> * once when read from the filesystem [collection:57](https://github.com/ethersphere/bee-js/pull/35/files#diff-e49f543a389f6fde536d9872dbd29d7373fd6186219fa95dacf879817b0309ddR67)
> * second time when creating in memory tar [tar:6-12](https://github.com/ethersphere/bee-js/pull/35/files#diff-e24585393060b7edf90d6527e7c9b16e681c030919a4b46de4a175ee36b8a367R6-R11)
> 
> I wonder if we need to buffer the file already in the collections.

This is a good point. So in theory `Buffer` should work everywhere where we need `Uint8Array` however if I didn't add the conversion the tests fail for both node and dom with different errors. From a quick investigation I found that the `tar-js` library   checks for Uint8Array incorrectly:
https://github.com/beatgammit/tar-js/blob/master/lib/tar.js#L31

I will investigate further.

src/modules/collection.ts
```diff
@@ -94,16 +74,41 @@ export async function buildCollection(dir: string, recursive = true): Promise<Co
...
+    fr.onload = () => resolve(fr.result as ArrayBuffer)
+    fr.readAsArrayBuffer(file)
+  })
+}
+
+interface WebkitFile extends File {
+  webkitRelativePath?: string
+}
+
+function filePath(file: WebkitFile) {
```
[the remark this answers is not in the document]
agazso: Here the WebkitFile is just for fixing the type definitions to better reflect reality. The `webkitRelativePath` is provided on files if you specify the `webkitdirectory` property on the HTML `input` element (this is a non-standard functionality supported in all major browsers). So we are not adding any new functionality just augment the type definition so that the compiler don't complain.

src/modules/collection.ts
agazso: See my comment for your other suggestion.

vojtechsimetka:
Well I was rather wondering whether we can not keep it as promisses until we put it into tar and ideally right away garbage collect. So we would process the files in loop, read it into `Uint8Array`, put into `tar-js` and remove from memory. For each file.

src/modules/collection.ts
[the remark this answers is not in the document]
nugaon: but why we can't overload that name getter instead?

src/utils/headers.ts
```diff
@@ -8,10 +6,10 @@ function readContentDispositionFilename(header?: string): string {
     if (!header) {
       throw new BeeError('missing content-disposition header')
     }
-    const disposition = contentDisposition.parse(header)
+    const dispositionMatch = header.match(/filename\*?=['"]?(?:UTF-\d['"]*)?([^;\r\n"']*)['"]?;?/i)
```
nugaon, line 9: I'm not the best to interpret regexes by looking at, but could you make a comment on what does it exactly do please?

src/utils/tar-js.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'tar-js'
```
nugaon, line 1: I suggest to place d.ts files into the types/ folder instead

src/utils/headers.ts
nugaon: maybe we should put a jsdoc on the function also in here.

agazso:
> Well I was rather wondering whether we can not keep it as promisses until we put it into tar and ideally right away garbage collect. So we would process the files in loop, read it into `Uint8Array`, put into `tar-js` and remove from memory. For each file.

I created a branch for that, you can check out:
https://github.com/ethersphere/bee-js/pull/43/files

I don't like it too much because it exposes the internals to the public API (by having `Buffer` and `ArrayBuffer` types) and considering the in-memory tar is just a workaround until we either implement the tar streaming in the browser or uploading individual files and add them to a manifest (when the API supports that).

src/modules/collection.ts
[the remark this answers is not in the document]
agazso: As I wrote before this is just a fix for the type definitions so we are not changing anything in the runtime behavior, that comes from the browser.

The other reason is that (as far as I know) it's not possible to define getters on interfaces.

But since this confused both of you I will add a comment explaining what's going here.

src/utils/headers.ts
agazso: Good idea

src/utils/tar-js.d.ts
agazso: Good idea

Issue #37, opened 2020-12-11
Readme: add bee configuration for local tests


vojtechsimetka:
Requested in https://github.com/ethersphere/bee-js/pull/25#pullrequestreview-546446385
> Provide a standard configuration for running bee locally for the tests

PR #41, opened 2020-12-12
chore: add Bee client starter with Docker

nugaon:
In order to have unified test results, we need the same Bee client configuration to do tests.

This PR adds a little Bash script, which starts a Bee client with the appropriate configurations for run local tests.

## TODO
- [x] - Amend `README` about this

extends #37

### DIFF
--- a/test/bee.sh
+++ b/test/bee.sh
@@ -1,21 +1,28 @@
 #!/bin/sh
-if [ ! -d /tmp/bee ] ; then
-   mkdir -p /tmp/bee/data && \
-   chown $(id -u):999 -R /tmp/bee && \
-   chmod 770 -R /tmp/bee && \
-   echo "password" > /tmp/bee/password
+# --ephemeral -> create ephemeral container for bee-client. Data won't be persisted
+BEE_CONTAINER=`docker container ls -qaf name=bee-test`
+if [ -z $BEE_CONTAINER ] || [ "$1" = "--ephemeral" ] ; then
+  BEE_IMAGE="ethersphere/bee:0.4.0"
+  BEE_PASSWORD="password"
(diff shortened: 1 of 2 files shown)


test/bee.sh
```diff
@@ -0,0 +1,21 @@
...
+   echo "password" > /tmp/bee/password
+fi
+docker run $@ \
+  -p 1635:1635 \
+  -p 1634:1634 \
+  -p 1633:1633 \
+  --volume /tmp/bee:/bee \
+  --interactive \
+  --tty \
+  "ethersphere/bee:0.4.0" \
```
vojtechsimetka, line 15: Could we use something that would just point to latest version? E.g. `ethersphere/bee`

vojtechsimetka (commented):
I agree we should provide this, but let's also review tests and lets put some to integration tests (e.g. tags).

test/bee.sh
nugaon: I think if we show which version we tested exactly is better.

nugaon:
@vojtechsimetka aggree, feel free and commit to this PR : )

test/bee.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+if [ ! -d /tmp/bee ] ; then
+   mkdir -p /tmp/bee/data && \
```
agazso, line 3: Won't this keep the data between runs? Maybe for tests it would be better to start from a clean state.

agazso (approved):

test/bee.sh
nugaon: Good idea, it keeps that, maybe even better to add an option to clear the prev. state of the client on demand

Issue #42, opened 2020-12-13
Review browser tests


nugaon:
With the current configuration is possible to test in headless Chromium browser, but the don't utilize it in our Jest tests.

We should [review our browser tests according to that](https://github.com/ethersphere/bee-js/pull/41#pullrequestreview-550863568).

agazso:
It seems that there is a problem with Typescript definitions when running browser tests. For example I created a test that is using the browser File API (that is defined in `lib.dom.d.ts` provided by Typescript) but when running the tests this gives a `ReferenceError: File is not defined` error message.

See:
https://github.com/ethersphere/bee-js/pull/47

PR #45, opened 2020-12-14
feat: pinning for files and collections

agazso:
This PR adds the pinning functionality for files and collections. It doesn't add the functionality for bytes and chunks because they are not part of the JS API yet.

This means that once we add the chunks functionality then we will need to add pinning for it as well and it has more endpoint than just post and delete. Regarding bytes: we may completely avoid to add it because currently it's not very different from files without a filename.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,8 @@
 import type { Readable } from 'stream'
 import * as file from './modules/file'
-import * as tag from './modules/tag'
 import * as collection from './modules/collection'
+import * as tag from './modules/tag'
+import * as pinning from './modules/pinning'
 import { Tag, FileData, Reference } from './types'
 
 /**
@@ -121,5 +122,41 @@ export default class Bee {
    */
   retrieveTag(tagUid: number | Tag): Promise<Tag> {
     return tag.retrieveTag(this.url, tagUid)
+  }
+
+  /**
+   * Pin file with given reference
+   *
+   * @param reference Bee file reference
+   */
+  pinFile(reference: Reference): Promise<pinning.Response> {
+    return pinning.pinFile(this.url, reference)
+  }
+
+  /**
+   * Unpin file with given reference
+   *
+   * @param reference Bee file reference
+   */
+  unpinFile(reference: Reference): Promise<pinning.Response> {
+    return pinning.unpinFile(this.url, reference)
+  }
+
+  /**
+   * Pin collection with given reference
+   *
+   * @param reference Bee collection reference
+   */
+  pinCollection(reference: Reference): Promise<pinning.Response> {
+    return pinning.pinCollection(this.url, reference)
+  }
+
+  /**
+   * Unpin collection with given reference
+   *
+   * @param reference Bee collection reference
+   */
+  unpinCollection(reference: Reference): Promise<pinning.Response> {
+    return pinning.unpinCollection(this.url, reference)
   }
 }
--- a/test/index.spec.ts
+++ b/test/index.spec.ts
@@ -23,4 +23,19 @@ describe('Bee class', () => {
 
     expect(tag).toEqual(tag2)
   })
+
+  it('should pin and unping files', async () => {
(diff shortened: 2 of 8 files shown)


src/index.ts
```diff
@@ -1,7 +1,8 @@
 import type { Readable } from 'stream'
 import * as file from './modules/file'
-import * as tag from './modules/tag'
 import * as collection from './modules/collection'
+import * as tag from './modules/tag'
+import * as pinning from './modules/pinning'
```
nugaon, line 5: imports that are not functions could be uppercase to denote it has properties <-> anyway this import style is not coherent with [test/modules/file.spec.ts](https://github.com/ethersphere/bee-js/pull/45/files#diff-20f72912331d5cd9ae8e7f007a50261131973bb0d20545176172f24d516414a1)

src/index.ts
```diff
@@ -122,4 +123,40 @@ export default class Bee {
...
+  }
+
+  /**
+   * Unpin collection with given reference
+   *
+   * @param reference Bee collection reference
+   */
+  unpinCollection(reference: Reference): Promise<pinning.Response> {
+    return pinning.unpinCollection(this.url, reference)
+  }
```
nugaon, line 161: Instead of `Promise<pinning.Response>` would be better to use `ReturnType<typeof pinning.[PINNING_METHOD_NAME]>`. It is better, because if its return type changes you don't have to change that here, also clearer what it exactly does (only wraps).

test/index.spec.ts
```diff
@@ -23,4 +23,19 @@ describe('Bee class', () => {
...
+    const content = new Uint8Array([1, 2, 3])
+
+    const hash = await bee.uploadFile(content)
+    await bee.pinFile(hash)
+    await bee.unpinFile(hash)
+  })
+
+  it('should pin and unping collection', async () => {
+    const path = './test/data/'
+    const hash = await bee.uploadFilesFromDirectory(path)
```
nugaon, line 37: Also it's available to write test to browser cases

test/modules/pinning.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  const invalidReference = '0000000000000000000000000000000000000000000000000000000000000000'
+
+  describe('should work with files', () => {
+    const randomData = randomByteArray(5000)
+
+    it('should pin an existing file', async () => {
+      const hash = await file.upload(BEE_URL, randomData)
+      const response = await pinning.pinFile(BEE_URL, hash)
+
+      expect(response.code).toBe(200)
```
nugaon, line 19: I suggest later to write own matchers to print out the error message of the bee-client, not only its http error code

nugaon (changes requested):

src/index.ts
agazso: The reason I started using lowercase imports was that the module names were conflicting types defined by us (`Collection`, `Tag`) or by builtin types (`File`). I will change `test/modules/file.spec.ts` to be consistent and open an issue about this.

https://github.com/ethersphere/bee-js/issues/46

src/index.ts
agazso: Generally this is a good idea however this here is the public interface of the library that we are providing. In that case I wouldn't expose an in implementation detail (e.g. the return type of the internally called function). Also I am not sure how that would look like in a generated documentation.

test/index.spec.ts
agazso: I tried to write a browser test, but I ran into an issue:

https://github.com/ethersphere/bee-js/issues/42#issuecomment-745230207

test/modules/pinning.spec.ts
agazso: Good idea, for now I added an `okResponse` variable and use that for all tests for comparison.

src/index.ts
nugaon: Alright then we can check it later

nugaon (approved):

PR #47, opened 2020-12-15
test: browser file test

agazso:
Failing browser test because lack of typing


nugaon (commented):
yeah, we should extend the global types with browser specific types, but you still don't call the functions from the browser context by this.

nugaon:
Anyway I started a branch related to #42 which make examples and archi for the browser tests

PR #49, opened 2020-12-16
feat: pss send and subscribe

agazso:
This PR implements the PSS send and subscribe functionality. There is also a `pssReceive` function on the Bee class to simplify the case when you want to receive just one message and it uses `pssSubscribe` internally. The WebSocket interface is not exposed on the public interface because I find it too cumbersome and error-prone to use. Instead a `PssMessageHandler` type is defined for recieving messages and error and a `PssSubscription` type is defined to provide cancel functionality on the subscription.

For this functionality we also need to access the debug API to get access to the overlay address and pssPublicKey. I added a DebugBee class temporarily to support this however I expect that on the long term these will move to a private API and its configuration may effect how we construct the Bee object as well.

The naming in the debug API is a bit inconsistent, I opened an issue about that: https://github.com/ethersphere/bee/issues/1078

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -289,6 +289,16 @@ export class Bee {
   /**
    * Receive message with Postal Service for Swarm
    *
+   * Because sending a PSS message is slow and CPU intensive,
+   * it is not supposed to be used for general messaging but
+   * most likely for setting up an encrypted communication
+   * channel by sending a one-off message.
+   *
+   * This is a helper function to wait for exactly one message to
+   * arrive and then cancel the subscription. Additionally a
+   * timeout can be provided for the message to arrive or else
+   * an error will be thrown.
+   *
    * @param topic Topic name
    * @param timeoutMsec Timeout in milliseconds
    *
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,7 +5,19 @@ import * as tag from './modules/tag'
 import * as pinning from './modules/pinning'
 import * as bytes from './modules/bytes'
 import * as pss from './modules/pss'
-import { Tag, FileData, Reference, UploadOptions, PublicKey } from './types'
+import * as connectivity from './modules/debug/connectivity'
+import {
+  Tag,
+  FileData,
+  Reference,
+  UploadOptions,
+  PublicKey,
+  AddressPrefix,
+  Address,
+  PssMessageHandler,
+  PssSubscription,
+} from './types'
+import { BeeError } from './utils/error'
 
 /**
  * The Bee class provides a way of interacting with the Bee APIs based on the provided url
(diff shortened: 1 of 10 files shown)


src/index.ts
```diff
@@ -209,6 +222,124 @@ export class Bee {
...
+      if (timeoutMsec > 0) {
+        // we need to cast the type because Typescript is getting confused with Node.js'
+        // alternative type definitions
+        timeout = (setTimeout(() => {
+          subscription.cancel()
+          reject(new BeeError('pssReceive timeout'))
+        }, timeoutMsec) as unknown) as number
+      }
+    })
+  }
```
vojtechsimetka, line 322: I don't quite understand why this method exists. What is the usecase tohave a method that subscribes and terminates after first message received?

src/index.ts
```diff
@@ -209,6 +222,124 @@ export class Bee {
...
+    }
+    ws.onerror = ev => {
+      // ignore errors after subscription was cancelled
+      if (!cancelled) {
+        handler.onError(new BeeError(ev.message), subscription)
+      }
+    }
+
+    return subscription
+  }
```
vojtechsimetka, line 287: I think passing in a handler is quite restrictive and it would be much better to return an instance of event emitter. That would allow us to have multiple subscribers with just one open websocket connection.

And I would even suggest we use [emittery](https://www.npmjs.com/package/emittery) which handles synchronicity way better than the standard event emitter and can await on event completion.

src/modules/debug/connectivity.ts
```diff
@@ -0,0 +1,34 @@
+import { safeAxios } from '../../utils/safeAxios'
+
+interface NodeAddresses {
+  overlay: string
+  underlay: string[]
+  ethereum: string
+  public_key: string
+  pss_public_key: string
+}
```
vojtechsimetka, line 9: Please extract the interface or at least export it.

src/modules/debug/connectivity.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  return response.data
+}
+
+interface Peer {
+  address: string
+}
+interface Peers {
+  peers: Peer[]
+}
```
vojtechsimetka, line 25: Please extract the interfaces or at least export them.

src/modules/pss.ts
```diff
@@ -0,0 +1,50 @@
...
+  return response.data
+}
+
+/**
+ * Subscribe for messages on the given topic
+ *
+ * @param topic Topic name
+ */
+export function subscribe(url: string, topic: string): WebSocket {
+  const wsUrl = url.replace(/^http/i, 'ws')
```
vojtechsimetka, line 47: I don't think this should be here. The URL that is passed should already be websocket address. Also IMHO the Bee class should in the constructor have option to specify the websocket address and if not provided, it is extracted from the http one (and vice versa)

vojtechsimetka: Or maybe PSS should actually be a standalone class from Bee. I assume many people will use one or the other but not both and their functionality is quite different.

vojtechsimetka (changes requested):

src/index.ts
agazso: The use-case is that because sending a PSS message is slow and CPU intensive, it is not supposed to be used for general messaging but most likely for setting up an encrypted communication channel by sending a one-off message.

For example on my machine sending one PSS message is between 2 and 20 seconds with more than 300% CPU utilization (on macOS). Currently the difficulty is set to mined chunk matches the first four hexdigits of the target but this may be longer in the future as the network grows, therefore the time to mine a chunk as well.

Quoting the Book of Swarm (pg. 124): 

> A small delay in the second range is expected only in a network of a billion nodes and even that is acceptable given that Trojan messages are meant to be used only for one-off instances such as initiations of a channel. All subsequent real-time exchange will happen using the previously described bidirectional outbox model using single owner chunks.

This is why I thought it would be good to have a function on the interface that does this and it's easy to use. While it's true that this can be done with `pssSubscribe` (as it is demonstrated by the implementation) I found that it's easy to forget about canceling the subscriptions correctly especially if you want to have some kind of timeout. Actually the first version of this function used `pss.subscribe` and the WebSocket API directly but I wanted to make sure that it can be implemented correctly with `pssSubscribe` to kind of "prove" that it has a usable interface I changed it.

I'm not insisting on having this function but if we aim our library to be easy-to-use (not just providing access to functionality of Bee) then I think this definitely belongs to here.

vojtechsimetka: Right, that makes sense. So basically we are deliberately suggesting way PSS should be used which is fine. I would then like to have the reasoning as a comment to the function. In general, I think we should focus a bit more one the comments quality (myself included! )

agazso: That's a very good point, I will add it to the comment.

src/index.ts
agazso: I don't like the handler too much either but I found that this was the easiest to use in a type-safe way without having to resort to an external library.

You can still have multiple subscribers if you want, for example with `emittery`:

```typescript
const emittery = new Emittery()
const subcription = bee.pssSubscribe(topic, {
  onMessage: message => emittery.emit('message', message),
  onError: error => emittery.emit('error', error),
})
```

The thinking behind using the handler is that (at least in Typescript) it enforces to specify the callbacks on the handler so it is correct by construction, whereas if we put an emitter on the return value then it cannot be guaranteed that the user will use it correctly.

I understand that this makes it quite opinionated and maybe one can argue that it's less idiomatic Javascript than using an emitter. But there is an other issue with using an emitter and that it's originally a Node.js class and (without an external package) it's not in the browser.

So we would need to use an external package to use that and I'm not against using one. My concern is with `emittery` that it's a 35kB package which would make the current bundle size 30% bigger and (if I understand correctly) it provides features that we wouldn't need.

Maybe a better alternative would to have a `bee-pss-emittery` package that provides the wrapping above for those who need it?

src/modules/pss.ts
agazso: I agree, this is a temporary workaround until we figure out how to separate the private and public APIs. I think the PSS endpoints should be in a private API (the sending because of the resource usage, the receiving because of that it's tied to a node) along with the node address query endpoint (which is currently in the debug API)

Until that point I wouldn't bother because with this the Bee class works as expected and that is what we expose (not the internal pss module). I think it's worth considering splitting the APIs in the future but then it would be still good to have one that exposes all the functionalities. For example as I wrote in my other comment the anticipated usage of PSS would be to set up an encrypted channel for bidirectional communication and that would be most likely implemented with single owner chunks.

src/modules/debug/connectivity.ts
agazso: Done

src/modules/debug/connectivity.ts
agazso: Done

src/index.ts
vojtechsimetka: Valid arguments, you convinced me. We can see what is the demand from users later on but right now this is fine :) Thanks!

src/modules/pss.ts
vojtechsimetka: Agree

vojtechsimetka (approved):
LGTM, thanks @agazso !

PR #50, opened 2020-12-16
feat: bytes module

agazso:
Add `/bytes` endpoint wrappers.

### DIFF
--- a/src/modules/pinning.ts
+++ b/src/modules/pinning.ts
@@ -9,6 +9,25 @@ export interface Response {
   code: number
 }
 
+async function pinRequest(url: string, method: 'post' | 'delete'): Promise<Response> {
+  const response = await safeAxios<Response>({
+    method,
+    responseType: 'json',
+    url,
+  })
+
+  return response.data
+
+}
+
+async function pin(url: string, endpoint: string, hash: string): Promise<Response> {
+  return pinRequest(`${url}${endpoint}/${hash}`, 'post')
+}
+
+async function unpin(url: string, endpoint: string, hash: string): Promise<Response> {
+  return pinRequest(`${url}${endpoint}/${hash}`, 'delete')
+}
+
 /**
  * Pin file with given reference
  *
@@ -16,13 +35,7 @@ export interface Response {
  * @param hash Bee file reference
  */
 export async function pinFile(url: string, hash: string): Promise<Response> {
-  const response = await safeAxios<Response>({
-    method: 'post',
-    responseType: 'json',
-    url: `${url}${fileEndpoint}/${hash}`,
-  })
-
-  return response.data
+  return pin(url, fileEndpoint, hash)
 }
 
 /**
@@ -32,13 +45,7 @@ export async function pinFile(url: string, hash: string): Promise<Response> {
  * @param hash Bee file reference
  */
 export async function unpinFile(url: string, hash: string): Promise<Response> {
-  const response = await safeAxios<Response>({
-    method: 'delete',
-    responseType: 'json',
-    url: `${url}${fileEndpoint}/${hash}`,
-  })
-
-  return response.data
+  return unpin(url, fileEndpoint, hash)
 }
 
 /**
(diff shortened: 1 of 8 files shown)


src/modules/pinning.ts
```diff
@@ -71,3 +72,35 @@ export async function unpinCollection(url: string, hash: string): Promise<Respon
...
+ */
+export async function unpinData(url: string, hash: string): Promise<Response> {
+  const response = await safeAxios<Response>({
+    method: 'delete',
+    responseType: 'json',
+    url: `${url}${bytesEndpoint}/${hash}`,
+  })
+
+  return response.data
+}
```
nugaon, line 106: (un)pinCollection, (un)pinFile and (un)pinData are literally the same. it should be one function with some enum parameter, which decides what endpoint should be called.

test/index.spec.ts
```diff
@@ -5,37 +5,68 @@ describe('Bee class', () => {
...
+    }
 
-    const hash = await bee.uploadFile(content)
-    await bee.pinFile(hash)
-    await bee.unpinFile(hash)
-  })
+    it('should pin and unping files', async () => {
+      const content = new Uint8Array([1, 2, 3])
+
+      const hash = await bee.uploadFile(content)
```
nugaon, line 40: most of the tests upload this same content. Maybe It could be better to place it in `beforeAll` hook func.

test/modules/bytes.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+    const data = 'hello world'
+
+    const hash = await bytes.upload(BEE_URL, data)
+    const downloadedData = await bytes.download(BEE_URL, hash)
+
+    expect(Buffer.from(downloadedData).toString()).toEqual(data)
+  })
+
+  it('should catch error', async () => {
+    const invalidReference = '0000000000000000000000000000000000000000000000000000000000000000'
```
nugaon, line 17: this would be better in `utils.ts` file as an exported constant as we use it regurarly.

nugaon, line 17: this would be better in `utils.ts` file as an exported constant as we use it regurarly.

test/modules/pinning.spec.ts
```diff
@@ -73,4 +74,30 @@ describe('modules/pin', () => {
       await expect(pinning.unpinCollection(BEE_URL, invalidReference)).rejects.toThrow('Not Found')
     })
   })
+
+  describe('should work with data', () => {
+    const randomData = randomByteArray(5000)
+
+    it('should pin existing data', async () => {
+      const hash = await bytes.upload(BEE_URL, randomData)
+      const response = await pinning.pinData(BEE_URL, hash)
```
nugaon, line 83: same as I mentioned at the other test file. From the perspective of the test, it's not necessary to upload at any cases

nugaon (changes requested):

src/modules/pinning.ts
agazso: Good idea

test/index.spec.ts
agazso: They upload the same content but because they use different functions the resulting hash will be different. Files are basically bytes + metadata, and manifests add even more metadata to store the paths.

test/modules/pinning.spec.ts
agazso: See my other comment regarding this

src/modules/pinning.ts
```diff
@@ -2,26 +2,39 @@ import { safeAxios } from '../utils/safeAxios'
...
+
+function pin(url: string, endpoint: string, hash: string): Promise<Response> {
+  return pinRequest(`${url}${endpoint}/${hash}`, 'post')
+}
+
+function unpin(url: string, endpoint: string, hash: string): Promise<Response> {
+  return pinRequest(`${url}${endpoint}/${hash}`, 'delete')
+}
```
nugaon, line 28: The whole file is more clear now, although I would change the `endpoint` to an enum typed or literal typed parameter, like `pinnableStructure: 'collection' | 'file' | 'bytes'` and let these functions construct the URLs based on that parameter. Thereby, you don't need those wrapping functions and the code is more transparent and maintainable.

nugaon (changes requested):

src/modules/pinning.ts
agazso: Changed the endpoint parameter on pin and unpin to enums.

nugaon (approved):

PR #51, opened 2020-12-17
test: real browser

nugaon:
I made a new compile command `npm run compile:window` which makes a bundle for testing in real browsers.

The code that it generates is the same as the already existing ones, except it assignes a initialized Bee class to the global object 
`window` in browsers. Thereby It is possible to run test in Jest within browser context, or do manual tests in the real browser itself by opening [test/testpage/testpage.html](https://github.com/ethersphere/bee-js/pull/51/files#diff-9fa65e10d418cd2ec709ec665ffb62ecae0bf353a37149caae256fd277d4e5e9). 

I've created examples how to write Jest test with Puppeteer.

## TODO
- [x] amend README about it

it resolves issue #42 
It closes PR #47


test/testpage/testpage.ts
```diff
@@ -1,3 +1,5 @@
 import { Bee } from '../../src'
 
+window.beeFactory = (...params: ConstructorParameters<typeof Bee>) => new Bee(...params)
```
agazso, line 3: Maybe I miscommuncited but the idea was to make Bee available in the browser context (for example by having `window.Bee = Bee`) and then in make an instance in the tests, like how it is used in `index.spec.ts`. Then you don't need to inject __BEE_URL__ but you can just simply use the `beeUrl` function from `utils` similarly to the other tests.

agazso (changes requested):

test/testpage/testpage.ts
nugaon: beeFactory provides you Bee instance with any arbitrary config, which basically the same thing that you would do with the direct binding. It's not necessary to make an instance in tests, because you can use the default `window.bee` object for most of the cases.
Unfortunately the `beeUrl` function in utils cannot be called from browser context, because it calls `process`, which is a NodeJS global.

nugaon: But I see your point, anyway it works well now and you even don't have to initiailze that at every tests.

agazso: Ok, let's keep it this way for now

agazso (approved):

PR #52, opened 2020-12-17
test: change jest test timeout to 15sec

agazso:
Don't merge: this is a test for checking if the test flakyness could be solved by increasing jest timeout to 15secs.


agazso:
Ultimately solved by reversing to Bee 0.4.1

It is possible that this commit has changed the chunk pinning behavior
https://github.com/ethersphere/bee/commit/f365a29a27b3ab9e10ac244b7f6b762cafe05332

PR #54, opened 2020-12-21
test: always use latest beeinfra

agazso:
Do not merge!

This is only for testing the latest bee version for regression regarding the pinning/unpinning timeouts.


vojtechsimetka:
Do we still need this?

agazso:
No, I will close it

PR #58, opened 2020-12-29
build: generate declaration files via tsc

nugaon:
It adds `npm run compile:types` command to generate declaration files for each ts file in folder `/src` 

Output dir is `dts`.


tsconfig.d.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "./tsconfig",
+  "compilerOptions": {
+    "strictNullChecks": false,
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "outDir": "dts",
```
agazso, line 7: What is the reason to put the .d.ts file in `dts` and not `dist` like the other output files?

nugaon: I just separated these two to see what the processes' outputs are exactly, anyway I could put declaration files to the `dist` folder as well.

tsconfig.json
```diff
@@ -13,6 +13,7 @@
   },
   "exclude": [
       "dist",
+      "dts",
```
agazso, line 16: We don't need this anymore

agazso (changes requested):

tsconfig.json
nugaon: yeah, it escaped my attention. Thanks!

agazso (approved):

PR #60, opened 2021-01-05
feat: chunk API module

agazso:
This PR implements the endpoints related to chunks (chunk upload, download and pinning). It only adds at the module level because at the `Bee` class level I plan to add higher level functionality (e.g. Chunk interface instead of  `Uint8Array`s).

The upload API is quite strange that it already requires the hash to be calculated in order to be able to upload. Since this functionality was not yet added, I hardcoded some pre-calculated values in the upload and download test.

The chunk pinning API has a few extra functions compared to the other (bytes, fiels, collection) APIs and testing it is a bit difficult with only having hardcoded values because this way the tests affect each other. I plan to change that in the next iteration when we are able to create (pseudo-)random chunks, in this PR the goal of the tests are to provide coverage for the endpoint wrapper functions.

### DIFF
--- a/src/modules/chunk.ts
+++ b/src/modules/chunk.ts
@@ -1,0 +1,70 @@
+import type { Readable } from 'stream'
+import type { BeeResponse, UploadOptions } from '../types'
+import { prepareData } from '../utils/data'
+import { extractUploadHeaders } from '../utils/headers'
+import { safeAxios } from '../utils/safeAxios'
+
+const endpoint = '/chunks'
+
+/**
+ * Upload chunk to a Bee node
+ *
+ * The chunk data consists of 8 byte span and up to 4096 bytes of payload data.
+ * The span stores the length of the payload in uint64 little endian encoding.
+ * Upload expects the chuck data to be set accordingly.
+ *
+ * @param url     Bee URL
+ * @param hash    Chunk reference
+ * @param data    Chunk data to be uploaded
+ * @param options Aditional options like tag, encryption, pinning
+ */
+export async function upload(
+  url: string,
+  hash: string,
+  data: Uint8Array,
+  options?: UploadOptions,
+): Promise<BeeResponse> {
+  const response = await safeAxios<BeeResponse>({
+    method: 'post',
+    url: `${url}${endpoint}/${hash}`,
+    data: await prepareData(data),
+    headers: {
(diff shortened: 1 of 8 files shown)


src/modules/chunk.ts
```diff
@@ -0,0 +1,70 @@
...
+ * The chunk data consists of 8 byte span and up to 4096 bytes of payload data.
+ * The span stores the length of the payload in uint64 little endian encoding.
+ * Upload expects the chuck data to be set accordingly.
+ *
+ * @param url     Bee URL
+ * @param hash    Chunk reference
+ * @param data    Chunk data to be uploaded
+ * @param options Aditional options like tag, encryption, pinning
+ */
+export async function upload(
```
vojtechsimetka, line 21: Do you think it makes sense to add validation in this function as well? I am talking about checking the `span`, `length of the payload` and `hash`. Or would you add this to some utilities?

vojtechsimetka (approved):
LGTM!

src/modules/chunk.ts
```diff
@@ -0,0 +1,70 @@
...
+import { safeAxios } from '../utils/safeAxios'
+
+const endpoint = '/chunks'
+
+/**
+ * Upload chunk to a Bee node
+ *
+ * The chunk data consists of 8 byte span and up to 4096 bytes of payload data.
+ * The span stores the length of the payload in uint64 little endian encoding.
+ * Upload expects the chuck data to be set accordingly.
```
nugaon, line 14: typeo

nugaon (approved):
The only problem that I noticed, on local test with cluster nodes the chunk (and also file, byte) download attempt on not existing reference throws timeout error.
we should raise the default jest timeout because of that.

src/modules/chunk.ts
agazso: What do you mean exactly? :)

src/modules/chunk.ts
agazso: I found this API too low-level for practical usage and probably subject to change anyhow. Because of this I also wouldn't expose this API on the `Bee` interface. Instead I plan to implement utilities for creating correct chunks and calculate the BMT hash. We will need that to implement single owner chunks, but this is only necessary because there is no semantic API yet for handling those. Then I would expose only the SOC related functions on the `Bee` interface.

Issue #62, opened 2021-01-06
Jest local test timeout


nugaon:
Environment where multiple nodes participate, download attempts on not existing reference throw timeout error.
we should raise the default jest timeout because of that.

PR #64, opened 2021-01-06
fix: raise payment tolerance on tests

nugaon:
at issue #62 turned out the script which is provided to run local tests in cluster, after a while, run out of the balance.
it fixes that problem by adding the maximum possible amount of payment tolerance value for the `queen` node.

### DIFF
--- a/test/bee.sh
+++ b/test/bee.sh
@@ -138,7 +138,8 @@ if [ -z "$QUEEN_CONTAINER_IN_DOCKER" ] || $EPHEMERAL ; then
         --swap-enable=false \
         --debug-api-enable \
         --welcome-message="You have found the queen of the beehive..." \
-        --cors-allowed-origins="*"
+        --cors-allowed-origins="*" \
+        --payment-tolerance 2147483647
 else
     docker start "$QUEEN_CONTAINER_IN_DOCKER"
 fi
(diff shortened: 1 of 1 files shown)


test/bee.sh
```diff
@@ -138,7 +138,8 @@ if [ -z "$QUEEN_CONTAINER_IN_DOCKER" ] || $EPHEMERAL ; then
         --swap-enable=false \
         --debug-api-enable \
         --welcome-message="You have found the queen of the beehive..." \
-        --cors-allowed-origins="*"
+        --cors-allowed-origins="*" \
+        --payment-tolerance 2147483647
```
agazso, line 142: I would add the payment-tolerance to the workers as well because in our tests we issue requests to the queen therefore it is the one which will be in debt.

agazso (changes requested):

test/bee.sh
nugaon: I added, although in tests we only use the queen's API EP as I noticed

agazso: I don't see the commit, maybe you forgot to push?

Actually that's exactly the issue that we only use the queen's API endpoint. When you want to download a non-existent reference it will ask the peers (the workers) for looking up. That costs balance for the queen and eventually the queen's balance will be negative at their peers, so that's where the payment tolerance is more important to be increased.

agazso (approved):

PR #65, opened 2021-01-06
feat: bmt hasher

agazso:
Added rudimentary BMT hasher (see BOS. 2.2.2) that is required to use the current Chunk API. 

There is a test that uploads content addressed chunks with the Chunk API to verify that it works correctly for several values. Also added a utility function to create spans. I didn't want to deal with `BigInt` compatibility issues so right now the span size is limited to 2**32 (4GB). Since we currently don't support streaming upload it should not be an issue.


src/chunk/bmt.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+function bmtRootHash(payload: Uint8Array): Uint8Array {
+  if (payload.length > MAX_CHUNK_PAYLOAD_SIZE) {
+    throw new BeeArgumentError('invalid data length', payload)
+  }
+
+  // create an input buffer padded with zeros
+  let input = new Uint8Array([...payload, ...new Uint8Array(MAX_CHUNK_PAYLOAD_SIZE - payload.length)])
+  while (input.length > HASH_SIZE) {
```
nugaon, line 25: instead of `>` I would use `!==`, because it expresses better the goal of the iteration. Anyway it also works fine.

test/chunk/bmt.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { bmtHash } from '../../src/chunk/bmt'
+import { beeUrl, byteArrayToHex, okResponse } from '../utils'
+import * as chunk from '../../src/modules/chunk'
+import { makeSpan } from '../../src/chunk/span'
+
+describe('bmt', () => {
+  it('should produce correct BMT hash', () => {
+    const payload = new Uint8Array([1, 2, 3])
```
nugaon, line 8: This test input could be debatable in that sense it even doesn't fill one segment.
Of course, basically we hash this with zeros that still represent data in the operation, but the function could operate well in the first cycle and then regardless of the input data could hash the same things.

nugaon (changes requested):

src/chunk/span.ts
```diff
@@ -0,0 +1,21 @@
+import { BeeArgumentError } from '../../src/utils/error'
+
+export function makeSpan(length: number): Uint8Array {
+  if (length <= 0) {
+    throw new BeeArgumentError('invalid length for span', length)
+  }
+
+  // we limit the maximum span size in 32 bits to avoid BigInt compatibility issues
+  if (length > 2 ** 32 - 1) {
```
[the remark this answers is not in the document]
agazso: Added a constant `MAX_SPAN_LENGTH` along with the comment

src/chunk/bmt.ts
```diff
@@ -0,0 +1,38 @@
+import { keccak256 } from 'js-sha3'
+import { BeeArgumentError } from '../utils/error'
+
+const MAX_CHUNK_PAYLOAD_SIZE = 4096
+const SEGMENT_SIZE = 32
```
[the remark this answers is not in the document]
agazso: Added a constant `SEGMENT_PAIR_SIZE` that is the double of `SEGMENT_SIZE` and changed the comment to explain better what's going on.

src/chunk/bmt.ts
agazso: Fixed

test/chunk/bmt.spec.ts
agazso: There are three tests:
1. This is the first test and indeed it doesn't fill one segment. However we use this test data in multiple tests including the chunk upload test with the same hardcoded hash to make sure it produces the same result. If the filling with zeroes would not work correctly it would produce a different hash.
2. The second one tests with different sizes from 1 to 4096 doubling in each iteration.
3. The third test fills a complete 4096 bytes length chunk with random data

nugaon: with this 3rd test is fine, thanks!

nugaon (approved):

PR #69, opened 2021-01-13
feat: single owner chunks

agazso:
This PR adds functionality to work with different Bee chunk types, such as:
- content addressed chunks
- intermediate chunks
- single owner chunks

This functionality still hasn't been exposed to the Bee class (for that discussion see #71) because the PR is already quite big enough without that. The following new features are implemented:
- fixed and flexible length type safe and run-time verifiable byte array types that can be used to define binary types
- Ethereum compatible signing and recovery functionality. For that we use the [elliptic](https://www.npmjs.com/package/elliptic) plain JS package which has more dependencies that I'd like generally, but I used it before and it was relatively easy to make it output the same signature format that Bee uses. We may consider trying alternatives in the future.
- `Chunk` interface and function to create different types of chunks, with signing is done through the provided `Signer` interface


src/chunk/soc.ts
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * Intermediate chunks are used to store data when the size
+ * exceeds the maximum size of a single chunk.
+ * @see MAX_PAYLOAD_SIZE
+ * It encapsulates references to its children.
+ *
+ * @param spanLength  The length of the subtree below the intermediate chunk
+ * @param references  The references to children chunks
+ */
+export function makeIntermediateChunk(spanLength: number, references: SpanReference[]): Chunk {
```
vojtechsimetka, line 89: Why is this function in the `soc.ts`? Also it is not used anywhere

src/chunk/soc.ts
```diff
@@ -0,0 +1,245 @@
...
+function flexBytesAtOffset<Min extends number, Max extends number>(
+  offset: number,
+  _min: Min,
+  _max: Max,
+  data: Uint8Array,
+): FlexBytes<Min, Max> {
+  return data.slice(offset) as FlexBytes<Min, Max>
+}
+
+function makeSingleOwnerChunkFromData(data: ValidSingleOwnerChunkData, address: () => ChunkAddress): SingleOwnerChunk {
```
vojtechsimetka, line 204: Why is the second argument an empty-input function? It is also used inconsistently see:
https://github.com/ethersphere/bee-js/pull/69/files#diff-557d1bf0408614c16e32e921e18d5f4a3e6d184bfd271e429ab09dbf4ee93769R244

and 

https://github.com/ethersphere/bee-js/pull/69/files#diff-557d1bf0408614c16e32e921e18d5f4a3e6d184bfd271e429ab09dbf4ee93769R182

src/chunk/soc.ts
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * Verifies if the data is a valid single owner chunk
+ *
+ * @param data    The chunk data
+ * @param address The address of the single owner chunk
+ *
+ * @returns a single owner chunk or throws error
+ */
+export function verifySingleOwnerChunk(data: Uint8Array, address: ChunkAddress): SingleOwnerChunk {
```
vojtechsimetka, line 180: I find the name confusing because it does not take as an input a `SingleOwnerChunk`. It's more of a `makeSingleOwnerChunk` method as it takes the data and chunk address as input.

Also it is only used in tests, maybe it should be in `test/chunk/soc.ts`

src/chunk/soc.ts
```diff
@@ -0,0 +1,245 @@
...
+    data,
+    identifier,
+    signature,
+    span,
+    payload,
+    address,
+  }
+}
+
+export function singleOwnerChunkAddress(identifier: Identifier, address: EthAddress): ChunkAddress {
```
vojtechsimetka, line 220: I find `makeSOCaddress` shorter and more descriptive

src/utils/hex.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @param hex string input
+ */
+export function hexToByteArray(hex: HexString): number[] {
+  const hexWithoutPrefix = stripHexPrefix(hex)
+  const subStrings: string[] = []
+  for (let i = 0; i < hexWithoutPrefix.length; i += 2) {
+    subStrings.push(hexWithoutPrefix.substr(i, 2))
+  }
+
+  return subStrings.map(s => parseInt(s, 16))
```
vojtechsimetka, line 29:
```suggestion
  return hexWithoutPrefix.match(/.{1,2}/g).map(byte => parseInt(byte, 16))
```

src/utils/hex.ts
```diff
@@ -0,0 +1,79 @@
...
+export function stripHexPrefix<T extends string>(hex: T): T {
+  return hex.startsWith('0x') ? (hex.slice(2) as T) : hex
+}
+
+/**
+ * Converts a hex string to array of numbers
+ *
+ * @param hex string input
+ */
+export function hexToByteArray(hex: HexString): number[] {
```
vojtechsimetka, line 22: IMHO we should avoid using `number[]` and just use `Uint8Array`

src/chunk/soc.ts
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * Verifies if a chunk is a valid content addressed chunk
+ *
+ * @param data          The chunk data
+ * @param chunkAddress  The address of the chunk
+ *
+ * @returns a valid content addressed chunk or throws error
+ */
+export function verifyChunk(data: Uint8Array, chunkAddress: ChunkAddress): Chunk {
```
vojtechsimetka, line 136: I find the name confusing because it does not take as an input Chunk. It's more of a makeChunk method as it takes the data and chunk address as input.

Also you never use the return value except for tests. Unless I am overlooking anything, that to me is unnecessary computation and memory usage. (see https://github.com/ethersphere/bee-js/pull/69/files#diff-557d1bf0408614c16e32e921e18d5f4a3e6d184bfd271e429ab09dbf4ee93769R237)

src/chunk/soc.ts
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * Creates a content addressed chunk and verifies the payload size.
+ *
+ * @param payloadBytes the data to be stored in the chunk
+ */
+export function makeContentAddressedChunk(payloadBytes: Uint8Array): Chunk {
+  const span = makeSpan(payloadBytes.length)
+  const payload = verifyFlexBytes(MIN_PAYLOAD_SIZE, MAX_PAYLOAD_SIZE, payloadBytes)
+  const data = serializeBytes(span, payload) as ValidChunkData
```
vojtechsimetka, line 72: More consistent with other usage and less magic

```suggestion
  verifyFlexBytes(MIN_PAYLOAD_SIZE, MAX_PAYLOAD_SIZE, payloadBytes)
  const data = serializeBytes(span, payloadBytes) as ValidChunkData
```

vojtechsimetka (approved):
Thanks @agazso , this looks good great.

I have made few suggestions which I see as improvements. Feel free to reject them.

src/utils/hex.ts
nugaon: @vojtechsimetka thanks for the suggestion.
Can I ask if we use any regex, can we put a comment above the codeline what exactly it does in order to human earthly mortals are able to understand and read immediately like me? : D
seems more effective, although I could read @agazso's code better.

vojtechsimetka: Sure. This one just splits the input into array an of 2 character long strings.

`.` matches any character except for line terminator
`{1,2}` quantifier does a greedy match of maximum 2 consequentive characters as long as there are two. If the total length is odd, it also returns the last character (that is why the minimum in quantifier is `1`)

e.g.
```
'1234' =>  ['12', '34'] 
'12345' =>  ['12', '34', '5']
```

test/chunk/serialize.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    const a2 = new Uint8Array([2])
+    const a3 = new Uint8Array([3])
+    const expectedResult = new Uint8Array([1, 2, 3])
+
+    const result = serializeBytes(a1, a2, a3)
+
+    expect(result).toEqual(expectedResult)
+  })
+
+  it('serializes2', () => {
```
nugaon, line 15: instead of this name, I would say `serializes chunk data with span` or something more straightforward.

src/chunk/signer.ts
```diff
@@ -0,0 +1,125 @@
...
+  return publicKeyToAddress(recPubKey)
+}
+
+/**
+ * Creates a default singer object that can be used when the private
+ * key is known.
+ *
+ * @param privateKey The private key
+ */
+export function makeDefaultSigner(privateKey: PrivateKey): Signer {
```
nugaon, line 116: Why is this function is not exposed outside of the test environment as well?

src/utils/hex.ts
nugaon: cool, I checked it, I'm just saying we should comment it above the codeline in the source code.

src/chunk/soc.ts
```diff
@@ -0,0 +1,245 @@
...
+type ValidSingleOwnerChunkData = BrandedType<Uint8Array, 'ValidSingleOwnerChunkData'>
+
+type SpanReference = Bytes<32> | Bytes<64>
+
+/**
+ * Creates a content addressed chunk and verifies the payload size.
+ *
+ * @param payloadBytes the data to be stored in the chunk
+ */
+export function makeContentAddressedChunk(payloadBytes: Uint8Array): Chunk {
```
nugaon, line 69: this function could be placed in other file, because `soc` should consist only single owner chunk related functionalities.

src/utils/hex.ts
vojtechsimetka: > seems more effective, although I could read @agazso's code better.

I just checked and it's actually not more efficient, I actually just found it way more readable. 

So let's keep @agazso code. If we do want to optimize (eliminate one cycle) we can do the `parseInt` right away

```suggestion
  const subStrings: string[] = []
  for (let i = 0; i < hexWithoutPrefix.length; i += 2) {
    subStrings.push(parseInt(hexWithoutPrefix.substr(i, 2), 16))
  }

  return subStrings
```

src/utils/hex.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @param withPrefix  Provides '0x' prefix when true (default: false)
+ */
+export function byteArrayToHex(byteArray: number[] | Uint8Array, withPrefix = false): HexString {
+  const prefix = withPrefix ? '0x' : ''
+
+  return (prefix +
+    Array.from(byteArray, byte => {
+      return ('0' + (byte & 0xff).toString(16)).slice(-2)
+    }).join('')) as HexString
+}
```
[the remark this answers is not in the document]
agazso: Cool, it's indeed much more readable than the original.

src/chunk/soc.ts
agazso: Indeed it shouldn't be here and it's not used anywhere, because chunk splitting is not yet implemented. For now I will remove it.

src/chunk/soc.ts
agazso: Both the `ContentAddressedChunk` and the `SingleOwnerChunk` is designed to hold the raw chunk data and all the properties of them can be accessed through empty-input functions. `address` is one such property. The inconsistency comes from because there are two ways to construct a single owner chunk (from raw byte data or from constructing and signing a content addressed chunk) in the first case we already have the address, but in the second we don't so I though we can just pass the function that returns the address.
I can see how this can be seen as a source of confusion so I will just change it so that the address is passed in instead.

src/chunk/soc.ts
agazso: See my comment above why we have two ways to create a single owner chunk. This would be the function to use when you download the chunk data and you want to create a `SingleOwnerChunk` object (as opposed to creating one with `makeSingleOwnerChunk` before uploading it).
Would `verifySingleOwnerChunkData` a better name for it?

src/chunk/soc.ts
agazso: Again I followed the same logic here as with single owner chunks: `verify` checks the byte data that is downloaded and constructs a chunk object, `make` creates it from the data you want to upload. In the case of content addressed chunk this is quite simple, as you can see `verifyChunk` eventually invokes `makeChunk` after the verification.

This verification may be not strongly necessary when working with a trusted local Bee node. However it is necessary when you work with nodes that you cannot trust (e.g. gateways or when connecting to peer nodes).

The other reason to add this was to encode the invariants for chunks so that we are able to test if the creation of the chunks are correct.

Right now none of these functions are exposed, that's the next step, for that see #71

src/utils/hex.ts
agazso: I like this latest version

src/chunk/soc.ts
agazso: I will move the content addressed chunk related functionality to `cac.ts`

nugaon (approved):
LGTM!

PR #77, opened 2021-01-26
fix: big file upload

### LINKED ISSUE #75: Axios error on big file upload
I tried to upload a bigger file (32MB) calling `uploadFile` function of the Bee class cause safeAxios error: `Request body larger than maxBodyLength limit`.

agazso:
Fixes #75

### DIFF
--- a/src/utils/safeAxios.ts
+++ b/src/utils/safeAxios.ts
@@ -5,7 +5,11 @@ axios.defaults.adapter = require('axios/lib/adapters/http') // https://stackover
 
 export async function safeAxios<T>(config: AxiosRequestConfig): Promise<AxiosResponse<T>> {
   try {
-    const response = await axios(config)
+    const response = await axios({
+      maxContentLength: Infinity,
+      maxBodyLength: Infinity,
+      ...config,
+    })
 
     return response
   } catch (e) {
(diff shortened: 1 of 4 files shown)


src/utils/safeAxios.ts
```diff
@@ -5,7 +5,11 @@ axios.defaults.adapter = require('axios/lib/adapters/http') // https://stackover
 
 export async function safeAxios<T>(config: AxiosRequestConfig): Promise<AxiosResponse<T>> {
   try {
-    const response = await axios(config)
+    const response = await axios({
+      maxContentLength: Infinity,
+      maxBodyLength: Infinity,
+      ...config,
+    })
```
vojtechsimetka, line 12: How does having the `maxContentLength` and `maxBodyLength` here helps over passing it in the `upload` function?

vojtechsimetka (approved):

src/utils/safeAxios.ts
agazso: `safeAxios` is a wrapper that is called from everywhere where we use `axios`. So changing it here works for all endpoint, we don't have to change all individual endpoints. Please note that this also disables the limits for downloading  according to the [documentation](https://github.com/axios/axios).

I just vaguely understand what's the point of this feature in axios (and it seems it's node only) but I think it's better to not expose it as is to our users. If we have a use-case where there is a point of limiting it, we can always expose an option that matches the context better.

Issue #71, opened 2021-01-19
Exposing SOC and Feed APIs on the Bee class


agazso:
When working on the Single Owner Chunk PR (#69) I realized that there are multiple ways of exposing the SOC API (and the Feed API after that) and there are different tradeoffs with the different solutions. Here I would like to present the alternatives that I came up with to discuss these aforementioned tradeoffs and to hopefully come up with a better solution.

The root problem is that the interface of the SOC API is asymmetric when uploading and downloading regarding the need of a signer object. When uploading an SOC it has to be signed already, so an implementation of the `Signer` interface must be passed to the upload function. The implementation of the `Singer` may be complicated because it may involve integration  with 3rd party libraries and applications that are security critical because they must store a private key and on behalf of the user are supposed to sign data (transactions in the case of Ethereum and chunks in the case of Swarm). While there is a default singer implementation provided for convenience, it expects the privateKey to be passed in directly which may not be a good idea to encourage the users from a security perspective.

Also one may imagine many use-cases when an application only wants to download data and never upload, or at least never upload signed content. In these cases it would be better if the user weren't forced to provide a `Signer` object. However when downloading in order to calculate the chunk address we need the ethereum account address (which is provided by the `Signer` interface by the way).

With that said here are the alternatives I came up with so far:

1. Expose upload and download as class functions on Bee:

This would look like the following:
```typescript
uploadSingleOwnerChunk(signer: Singer, identifier: Identifier, data: Uint8Array, options?: UploadOptions): Promise<BeeResponse>
downloadSingleOwnerChunk(ownerAddress: EthAddress, identifier: Identifier): Promise<SingleOwnerChunk>
```

Pros:
- simplicity from the implementation standpoint, it is very similar how the other API endpoints are exposed
- download doesn't need the signer interface, only the ethereum address

Cons:
- the user is supposed to keep track of which address and identifier is used for uploading and downloading because there is no interface holding that information

2. Expose signed Bee class for providing operations that require signing

```typescript
class Bee {
  // ...details are omitted
  withSigner(signer: Signer): BeeWithSigner
}

class BeeWithSigner {
  uploadSOC(identifier: Identifier, data: Uint8Array, options?: UploadOptions): Promise<BeeResponse>
  downloadSOC(identifier: Identifier): Promise<SingleOwnerChunk>
}
```

Pros:
- the user only need to instantiate a `BeeWithSigner` (name is up for discussion) object and upload and download happens with the same owner automatically

Cons:
[...]

PR #80, opened 2021-01-27
feat: soc interface

agazso:
Proof of concept for exposing the SOC interface on the `Bee` class as discussed in #71.

Currently I see it as a good tradeoffs between the alternatives, maybe with better names.


agazso:
Implemented in https://github.com/ethersphere/bee-js/pull/147

PR #85, opened 2021-01-28
docs: generating API docs through npm script

vojtechsimetka:
This fixes the docs generation which has been broken for a while. I decided to split the autogeneration through CI into a separate PR which I think is not urgent and may end up in the [bee-js.github.io repo](https://github.com/ethersphere/bee-js.github.io) and not this one.


tsconfig.d.json
```diff
@@ -6,7 +6,7 @@
     "emitDeclarationOnly": true,
     "outDir": "dist",
     "rootDirs": ["src"],
-    "skipLibCheck": true
-  },
-  "include": ["src"]
+    "skipLibCheck": true,
+    "typeRoots": ["./types"]
```
agazso, line 10: Can you explain how does this work? As far as I see under `./types` there is only one file, the `global.d.ts` which contains definitions for the browser tests. Is that the correct one to use?

agazso (approved):
LGTM

tsconfig.d.json
vojtechsimetka: Yes that would be a mistake, let me try to make that an empty array.

The whole purpose was to set this to any directory containing relevant types for the library. If `typeRoots` is not set, by default, webpack drags in all the typings from `node_modules` and even global ones. That causes issues for tools like typedoc but could also end up poluting the typings of the library itself with some dependencie.

Issue #68, opened 2021-01-11
Add balance and chequebook related endpoints


agazso:
Balance: https://docs.ethswarm.org/debug-api/#tag/Balance
Chequebook: https://docs.ethswarm.org/debug-api/#tag/Chequebook
Settlements: https://docs.ethswarm.org/debug-api/#tag/Settlements

PR #96, opened 2021-02-01
feat: balances and consumed endpoint

agazso:
Part of #68 

Adds the `/balances` and `/consumed` endpoints described here: https://docs.ethswarm.org/debug-api/#tag/Balance


vojtechsimetka (approved):
In the tests I would like to see some deeper comparison of the return data - at least checking the interface.

Issue #59, opened 2021-01-05
Add configuration options to Bee class constructor


agazso:
When setting up the `Bee` class the constructor only requires a url to be passed. This is not always enough, for example we are using a convention for calculating the Debug API port address from the passed url but since this is user configurable in Bee this may be not always correct.

One idea to address this is that the constructor would accept a second optional argument of object type where all the optional features could be configured.

AuHau:
I think this is already present with in `Bee` class for default signer and I don't see any more usecases that we could bring into it, so for now I am closing this.

PR #97, opened 2021-02-02
feat: chequebook api

agazso:
Part of #68

This PR provides the wrapper functions around the chequebook endpoints. Not all of them have tests because it is not trivial to create the conditions required because of the stateful nature of the functionality.

Also it's not trivial to run the tests locally because they require at least one Bee node running with `swap-enable` set to true and our test runner script currently doesn't do that. The chequebook tests now check if the `BEE_CHEQUEBOOK_URL` environment variable is set and if it's not then those tests are skipped and a message is printed to the console informing about the situation. This environment variable is set when the tests are running on github because there the test runners have `swap-enable` set.

Please note that this functionality is not yet exposed on the Bee class. I plan to implement the [`settlements`](https://docs.ethswarm.org/debug-api/#tag/Settlements) endpoint functions in a different PR, then do the integration of the debug endpoints together (which is related to #59) in another PR.

See also #96 and #98

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,6 +13,7 @@ jobs:
     env:
       REPLICA: 5
       BEE_URL: 'http://bee-0.localhost'
+      BEE_CHEQUEBOOK_URL: 'http://bee-0-debug.localhost'
 
     runs-on: ubuntu-latest
 
--- a/src/modules/debug/chequebook.ts
+++ b/src/modules/debug/chequebook.ts
@@ -40,13 +40,13 @@ export async function getChequeubookBalance(url: string): Promise<ChequebookBala
   return response.data
 }
 
-interface CashoutResult {
+export interface CashoutResult {
   recipient: string
   lastPayout: number
   bounced: boolean
 }
 
-interface LastCashoutActionResponse {
+export interface LastCashoutActionResponse {
   peer: string
   chequebook: string
   cumulativePayout: number
@@ -70,7 +70,7 @@ export async function getLastCashoutAction(url: string, peer: string): Promise<L
   return response.data
 }
 
-interface CashoutResponse {
+export interface CashoutResponse {
   transactionHash: string
 }
 
@@ -90,13 +90,13 @@ export async function cashoutLastCheque(url: string, peer: string): Promise<Cash
   return response.data
 }
 
-interface Cheque {
+export interface Cheque {
   beneficiary: string
   chequebook: string
   payout: number
 }
 
-interface LastChequesForPeerResponse {
+export interface LastChequesForPeerResponse {
   peer: string
   lastreceived: Cheque
   lastsent: Cheque
@@ -117,7 +117,7 @@ export async function getLastChequesForPeer(url: string, peer: string): Promise<
   return response.data
 }
 
-interface LastChequesResponse {
+export interface LastChequesResponse {
   lastcheques: LastChequesForPeerResponse[]
 }
 
@@ -135,7 +135,7 @@ export async function getLastCheques(url: string): Promise<LastChequesResponse> 
   return response.data
 }
 
-interface DepositTokensResponse {
+export interface DepositTokensResponse {
   transactionHash: string
 }
 
@@ -156,7 +156,7 @@ export async function depositTokens(url: string, amount: number): Promise<Deposi
   return response.data
 }
 
-interface WithdrawTokensResponse {
+export interface WithdrawTokensResponse {
   transactionHash: string
 }
 
--- a/src/modules/debug/chequebook.ts
+++ b/src/modules/debug/chequebook.ts
@@ -3,7 +3,7 @@ import { safeAxios } from '../../utils/safeAxios'
 const chequebookEndpoint = '/chequebook'
 
 interface CheuqebookAddressResponse {
-  checkbookaddress: string
+  chequebookaddress: string
 }
 
 /**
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -13,6 +13,7 @@ jobs:
     env:
       REPLICA: 5
       BEE_URL: 'http://bee-0.localhost'
+      BEE_CHEQUEBOOK_URL: 'http://bee-0-debug.localhost'
```
vojtechsimetka, line 16: Shouldn't this be called `BEE_DEBUG_URL`? To stay consistent with the `bee` documentation plus it also includes the connectivity, chunk and status endpoints

src/modules/debug/chequebook.ts
```diff
@@ -0,0 +1,178 @@
...
+export async function getChequeubookBalance(url: string): Promise<ChequebookBalanceResponse> {
+  const response = await safeAxios<ChequebookBalanceResponse>({
+    url: url + chequebookEndpoint + '/balance',
+    responseType: 'json',
+  })
+
+  return response.data
+}
+
+interface CashoutResult {
```
vojtechsimetka, line 43:
```suggestion
export interface CashoutResult {
```

src/modules/debug/chequebook.ts
```diff
@@ -0,0 +1,178 @@
...
+  recipient: string
+  lastPayout: number
+  bounced: boolean
+}
+
+interface LastCashoutActionResponse {
```
vojtechsimetka, line 49:
```suggestion
export interface LastCashoutActionResponse {
```

src/modules/debug/chequebook.ts
```diff
@@ -0,0 +1,178 @@
...
+export async function getLastCashoutAction(url: string, peer: string): Promise<LastCashoutActionResponse> {
+  const response = await safeAxios<LastCashoutActionResponse>({
+    url: url + chequebookEndpoint + `/cashout/${peer}`,
+    responseType: 'json',
+  })
+
+  return response.data
+}
+
+interface CashoutResponse {
```
vojtechsimetka, line 73:
```suggestion
export interface CashoutResponse {
```

src/modules/debug/chequebook.ts
```diff
@@ -0,0 +1,178 @@
...
+  const response = await safeAxios<LastCashoutActionResponse>({
+    method: 'post',
+    url: url + chequebookEndpoint + `/cashout/${peer}`,
+    responseType: 'json',
+  })
+
+  return response.data
+}
+
+interface Cheque {
```
vojtechsimetka, line 93:
```suggestion
export interface Cheque {
```

src/modules/debug/chequebook.ts
```diff
@@ -0,0 +1,178 @@
...
+  beneficiary: string
+  chequebook: string
+  payout: number
+}
+
+interface LastChequesForPeerResponse {
```
vojtechsimetka, line 99:
```suggestion
export interface LastChequesForPeerResponse {
```

src/modules/debug/chequebook.ts
```diff
@@ -0,0 +1,178 @@
...
+export async function getLastChequesForPeer(url: string, peer: string): Promise<LastChequesForPeerResponse> {
+  const response = await safeAxios<LastChequesForPeerResponse>({
+    url: url + chequebookEndpoint + `/cheque/${peer}`,
+    responseType: 'json',
+  })
+
+  return response.data
+}
+
+interface LastChequesResponse {
```
vojtechsimetka, line 120:
```suggestion
export interface LastChequesResponse {
```

src/modules/debug/chequebook.ts
```diff
@@ -0,0 +1,178 @@
...
+export async function getLastCheques(url: string): Promise<LastChequesResponse> {
+  const response = await safeAxios<LastChequesResponse>({
+    url: url + chequebookEndpoint + '/cheque',
+    responseType: 'json',
+  })
+
+  return response.data
+}
+
+interface DepositTokensResponse {
```
vojtechsimetka, line 138:
```suggestion
export interface DepositTokensResponse {
```

src/modules/debug/chequebook.ts
```diff
@@ -0,0 +1,178 @@
...
+    method: 'post',
+    url: url + chequebookEndpoint + '/deposit',
+    responseType: 'json',
+    params: { amount },
+  })
+
+  return response.data
+}
+
+interface WithdrawTokensResponse {
```
vojtechsimetka, line 159:
```suggestion
export interface WithdrawTokensResponse {
```

vojtechsimetka (approved):
I think the interfaces should probably be in a different files, exported and we should import them in the `src/index.ts` so that users can use them in their projects when declaring stuff.

.github/workflows/ci.yaml
agazso: Not exactly, although I understand the source of confusion.

So `BEE_CHEQUEBOOK_URL` is needed if you want to run the chequebook tests. It has to point to a debug url of a node that has `swap-enable` set to true. In this sense `BEE_CHEQUEBOOK_URL` is always a debug url, but a debug url is not necessarily true the other way.

Maybe instead of `BEE_CHEQUEBOOK_URL` it should be called something else to make this clearer? I was also thinking `BEE_CHEQUEBOOK_DEBUG_URL`

PR #100, opened 2021-02-04
refactor: rename BEE_CHEQUEBOOK_URL to BEE_TEST_CHEQUEBOOK

agazso:
Before the environment variable had to set to a url of a node with `swap-enable`. Now we just check if the `BEE_TEST_CHEQUEBOOK` environment variable is set or not and run the tests on the debug url.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,7 +13,7 @@ jobs:
     env:
       REPLICA: 5
       BEE_URL: 'http://bee-0.localhost'
-      BEE_CHEQUEBOOK_URL: 'http://bee-0-debug.localhost'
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     env:
       REPLICA: 5
       BEE_URL: 'http://bee-0.localhost'
-      BEE_CHEQUEBOOK_URL: 'http://bee-0-debug.localhost'
+      BEE_TEST_CHEQUEBOOK: 'test'
```
vojtechsimetka, line 16: Any reason this is not just `true`?
```suggestion
      BEE_TEST_CHEQUEBOOK: true
```

vojtechsimetka (approved):

PR #105, opened 2021-02-04
refactor!: breaking api changes

agazso:
This PR contains fixes for the breaking changes for the upcoming 0.5.0 Bee release.
The main breaking changes:
- the `/tags` endpoint returns different values
- uploading a chunk to the `/chunks` endpoint now returns the reference
- there is a new `/soc` endpoint to upload single owner chunks


vojtechsimetka:
One suggestion. If this is breaking change on our interface, you may want to add `!` after the scope (in this case `refactor!: ...` ). This way it will be properly caught by the `please-release` system and it will be listed in the breaking changes.

src/chunk/upload.ts
```diff
@@ -12,9 +13,15 @@ import * as chunkAPI from '../modules/chunk'
  * @param chunk     A chunk object
  * @param options   Upload options
  */
-export function uploadChunk(url: string, chunk: Chunk, options?: UploadOptions): Promise<BeeResponse> {
-  const address = chunk.address()
-  const hash = bytesToHex(address)
+export function uploadSingleOwnerChunk(
```
nugaon, line 16: It is purely only a SOC function, that could be moved to `src/chunk/soc.ts`

nugaon (changes requested):

src/chunk/soc.ts
```diff
@@ -67,22 +51,36 @@ export function isValidSingleOwnerChunkData(
...
+    const identifier = verifyBytesAtOffset(SOC_IDENTIFIER_OFFSET, IDENTIFIER_SIZE, data)
+    const socAddress = keccak256Hash(identifier, ownerAddress)
+
+    if (!bytesEqual(address, socAddress)) {
+      throw new BeeError('verifySingleOwnerChunk')
+    }
+
+    return makeSingleOwnerChunkFromData(data, address, ownerAddress)
+  } catch (e) {
+    throw e
```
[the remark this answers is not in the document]
agazso: Good catch, I will remove the `try ... catch`

.github/workflows/ci.yaml
```diff
@@ -34,7 +34,8 @@ jobs:
           export URL=$(curl -s https://api.github.com/repos/ethersphere/bee-local/releases/latest | jq -r .tarball_url)
           curl -Ls ${URL} -o bee-local.tar.gz
           tar --strip-components=1 --wildcards -xzf bee-local.tar.gz ethersphere-bee-local-*/{beeinfra.sh,helm-values,hack}
-          sed -i 's/IMAGE_TAG="latest"/IMAGE_TAG="0.4.2"/' beeinfra.sh
+          # TODO remove comment and update version number once the release is out
+          # sed -i 's/IMAGE_TAG="latest"/IMAGE_TAG="0.4.2"/' beeinfra.sh
```
vojtechsimetka, line 38: Just so we don't forget about this :)
```suggestion
          sed -i 's/IMAGE_TAG="latest"/IMAGE_TAG="0.5.0"/' beeinfra.sh
```

vojtechsimetka (approved):

src/modules/debug/settlements.ts
```diff
@@ -4,7 +4,7 @@ const settlementsEndpoint = '/settlements'
 
 export interface Settlements {
   peer: string
-  received: number
+  received: number | null
```
nugaon, line 7: I would write `received?: number` instead of that

test/modules/debug/settlements.spec.ts
```diff
@@ -25,7 +25,9 @@
       const peerSettlementResponse = await settlements.getSettlements(beeDebugUrl(), peerSettlement.peer)
 
       expect(peerSettlementResponse.peer).toEqual(peerSettlement.peer)
-      expect(typeof peerSettlementResponse.received).toBe('number')
+      // FIXME this is due to a bug in Bee
```
nugaon, line 28: you commented 12 hours ago it is fixed in Bee PR [#1213](https://github.com/ethersphere/bee/pull/1213).
could you correct this?

nugaon (approved):

Issue #76, opened 2021-01-23
Minimize the browser bundle


vojtechsimetka:
Right now, we are only generating non-minimized version of the bundles. We should minimize it so that users can use it directly in their dapps as is when they don't want to use webpack.

PR #107, opened 2021-02-05
ci: minify bundle

nugaon:
Minifies bundle for both node and browser version on production compilation.
Also fixes compilation by adding 'utf8-validate' and 'bufferutil' packages to the dependency list.

Fixes: #76

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -48,14 +48,13 @@ jobs:
           CI: true
 
       - name: Build nodejs
-        run: npm run compile:node
+        run: npm run compile:node -- --env mode=production
 
       - name: Build browser
-        run: npm run compile:browser
+        run: npm run compile:browser -- --env mode=production
 
       - name: Build typings
         run: npm run compile:types
 
       - name: Build docs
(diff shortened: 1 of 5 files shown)


agazso:
> Also fixes compilation by adding 'utf8-validate' and 'bufferutil' packages to the dependency list.

I saw the warnings with webpack compilation regarding these, but I tried the bundle in both browser and node and it worked fine. Are you sure we need this as dependencies? Maybe it's enough to have it as devDependency?

nugaon:
> > Also fixes compilation by adding 'utf8-validate' and 'bufferutil' packages to the dependency list.
> 
> I saw the warnings with webpack compilation regarding these, but I tried the bundle in both browser and node and it worked fine. Are you sure we need this as dependencies? Maybe it's enough to have it as devDependency?

It seems enough if it belongs to `devDependencies`, you are right! I hope we only use these libs in tests tough.

vojtechsimetka (commented):
Can you also change in the README.md the link for the unpkg.com to point to the minified version?

agazso (approved):

PR #120, opened 2021-02-06
fix: typescript configuration

### LINKED ISSUE #119: VS Code does not load custom types
After merging #118 and a Visual Studio Code reload, custom defined types did not load again in the editor.

I noticed if I move folder `test/types` to the root project folder, then move back these types work again.

Somehow should be solved to load custom types after IDE reload as well.

nugaon:
In order to load custom types in Visual Studio Code, we need to remove property `typeRoots` completely from `ṫsconfig.json` and move all used type definitions to `types` folder. 
Even if it has only one element like `"typeRoots": ["types"]` or `typeRoots": ["./types"]` it still won't load.

As I checked, `./src/types` is not necessary to be defined in `typeRoots` anyway, currently it only affects browser tests.

Fixes #119


nugaon:
as I see on [the test tun](https://github.com/ethersphere/bee-js/pull/120/checks?check_run_id=1845936095), this `typeRoots` definition eventually affects doc generation.

@vojtechsimetka , could you tell me some info about it?

vojtechsimetka (changes requested):
@nugaon 
> as I see on the test tun, this typeRoots definition eventually affects doc generation.

Yep, that is exactly the issue. If the `typeRoots` is not specified, webpack (and typedoc) is dragging in the type definitions of all the imports including `node_modules` and even the global ones (similar issue happened [here](https://github.com/sveltejs/svelte/issues/5692#issuecomment-731469915) ). It took me quite some time to find that this is the problem typedoc has, maybe there is some workaround but I spent 2+ days trying to find various reasons.

IMHO having API documentation superseeds a type warning in VScode (especially since e.g. Atom does not have this issue). If you want to dive into debugging typedoc setup in our project please do, here are some pointers that might help :)

- [search on typedoc repo for typeRoots](https://github.com/TypeStrong/typedoc/issues?q=+typeRoots)
- everything works (and worked) with Tasegir [where typeroots are dynamically added](https://github.com/AuHau/tasegir/blob/master/src/config/tsconfig.js#L24-L36)
- [switch to some other tool that would use TSDoc standard](https://github.com/microsoft/tsdoc)

nugaon:
> IMHO having API documentation superseeds a type warning in VScode (especially since e.g. Atom does not have this issue).

I have a different opinion about this and also I would rather use both. The types are more important in code as we are programming in TypeScript. 

P.S. I would'nt like to change my IDE. :smile:

vojtechsimetka (approved):
Nice! Happy you don't have to change your IDE 😬

Issue #103, opened 2021-02-04
Export some of the utility functions


vojtechsimetka:
## Rationale
It is likely some of the common functions like `hexToBytes` could be useful for devs who use our library.

## Task
Export new namespace `utils` with some of the most sensible functions. E.g `hexToBytes`, `bytesToHex` or `stripPrefix`

nugaon:
I suggest using CamelCase with uppercase starting letter on namespaces (same as in case of objects, which only contain other properties).

See #46

PR #123, opened 2021-02-08
feat: export utils and types

### LINKED ISSUE #102: Export types from index.js
## Rationalle
TypeScript users will likely want to declare variables or return types of their functions which will come from bee-js package. Therefore we should export them in `src/index.ts`

## Task
Ensure all types on input arguments and return types of functions are exported from `src/index.ts`

nugaon:
Export utilities and types from `index`.

Also extend `window` object with `BeeJs`.

Later these could be provided via other packages like `@types/bee-js` and `@ethersphere/bee-js-utils` and removed from compiled code to reduce output file size.

_manual tests needed from reviewers_

Resolves #102, resolves: #103


vojtechsimetka:
@nugaon can you show to me, how to import and use anything from Types? I can only use the two constants with:

```ts
import { Types } from "@ethersphere/bee-js"

// Nothing outside of the following works
Types.ENCRYPTED_REFERENCE_LENGTH
Types.REFERENCE_LENGTH
```

nugaon:
> @nugaon can you show to me, how to import and use anything from Types?

like this:
```ts
import { Types } from '@ethersphere/bee-js'
const vmi: Types.FlavoredType<string, 'ownType'> = 'value'
```

nugaon:
> Also I would personally prefer for the types to be exported directly without this Types wrapper.

I uploaded this approach

src/bee.ts
```diff
@@ -0,0 +1,343 @@
...
+  FileData,
+  Reference,
+  UploadOptions,
+  PublicKey,
+  AddressPrefix,
+  PssMessageHandler,
+  PssSubscription,
+  BeeResponse,
+} from './types'
+import { BeeError } from './utils/error'
```
vojtechsimetka, line 19: I think this type is not exported.

src/index.ts
```diff
@@ -1,495 +1,18 @@
...
-    }
-
-    const subscription = {
-      topic,
-      cancel,
+import { Bee } from './bee'
+import { BeeDebug } from './bee-debug'
+
+export * as Utils from './utils'
+export * from './types/index'
```
vojtechsimetka, line 5: I wonder if we need to export all the types, but over-exporting is better than not exporting something :)

src/utils/index.ts
```diff
@@ -0,0 +1,10 @@
+import { safeAxios } from './safeAxios'
```
vojtechsimetka, line 1: This can be removed

agazso, line 1: Shouldn't this be an export as well?

src/utils/index.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export * as Bytes from './bytes'
+export * as Data from './data'
```
vojtechsimetka, line 4: Probably does not need to be exported

src/utils/index.ts
```diff
@@ -0,0 +1,10 @@
...
+export * as Error from './error'
```
vojtechsimetka, line 5: IMHO should be exported as types only

src/utils/index.ts
```diff
@@ -0,0 +1,10 @@
...
+export * as File from './file'
```
vojtechsimetka, line 6: Also not really user facing

src/utils/index.ts
```diff
@@ -0,0 +1,10 @@
...
+export * as Headers from './headers'
```
vojtechsimetka, line 7: Not really user facing

src/utils/index.ts
```diff
@@ -0,0 +1,10 @@
...
+export * as Hex from './hex'
+export * as SafeAxios from './safeAxios'
```
vojtechsimetka, line 9: IMHO just internal function

src/utils/index.ts
```diff
@@ -0,0 +1,10 @@
...
+export * as Tar from './tar'
```
vojtechsimetka, line 10: Also internal only

agazso, line 10: If I understand correctly everytime we add a new file under utils that we want to expose it had to be added here as well.

I think it would be good to write this down in a comment somewhere (for example on the top of this file).

src/utils/index.ts
nugaon: it is exported blow, it just left accidentally, I'm going to remove.

src/index.ts
nugaon: if one type shouldn't be exported, then shoudln't be exported by default or at least should be placed in different folder than this common one.

src/utils/index.ts
nugaon: I think this wrapper actually nice on this.
Also I would like to keep the structure clear as it is in file structure.

src/utils/index.ts
vojtechsimetka: I think we should better curate this lists probably directly in the `src/index.ts`. This `src/utils/index.ts` has a slightly different use than that (it helps import from directory without knowing the file) but probably should not be enduser facing

src/index.ts
vojtechsimetka: Just to stay consistent :)
```suggestion
export * from './types'
```
> if one type shouldn't be exported, then shouldn't be exported by default 

Yeah I could get behind this

src/utils/index.ts
vojtechsimetka: Cool, then I would argue it should also be exported as a type directly in the `src/index.ts`. So `export type * from './utils/error'`

src/utils/index.ts
nugaon: because of the `Hex` funtionalities we started this PR basically.
Regards to the others I don't think that is a problem if we export these under `Utils` wrapper, because these could be handy as well (for example `fileArrayBuffer`).

src/utils/index.ts
nugaon: I try to answer all of your comments regards to this `utils/index.ts`
because of the Hex funtionalities we started this PR basically.
Regards to the others I don't think that is a problem if we export these under Utils wrapper, because these could be handy as well (for example fileArrayBuffer).
I will rename this `index.ts` to `export.ts` because it will lose its meaning if we would like to change anything which currently has.

From that you mentioned I only would remove `safeAxios` an `headers` and export `Errors`' members directly (I guess the last one is nice for doc generation, am I right?).

src/index.ts
```diff
@@ -1,495 +1,18 @@
...
-    const subscription = {
-      topic,
-      cancel,
+import { Bee } from './bee'
+import { BeeDebug } from './bee-debug'
+
+export * as Utils from './utils/expose'
+export * from './types'
+export { Bee, BeeDebug }
+
```
vojtechsimetka, line 7: This is untested code suggestion
```suggestion
export type * from './utils/error'

```

vojtechsimetka (approved):

src/index.ts
nugaon: if we would like to export errors directly, then I take out from `Utils`

vojtechsimetka: I think we should export type of the error here, but the actual error (as instantiable class) should be in utils.

agazso (approved):

PR #124, opened 2021-02-08
refactor: change bee url env

nugaon:
### Description
* change `BEE_URL` to `BEE_API_URL` as it represents the API endpoint of the Bee client.
* change `BEE_PEER_URL` to `BEE_PEER_API_URL`

#### Other refactor ideas **for an other refactor branch**
* should be refactored the debug url handling (introduce `BEE_DEBUG_URL` or something similar),
* also seems a good idea to change default fallback values of these URL variables to a more straightforward one (e.g. `http://localhost:1633` on `BEE_API_URL` and `http://localhost:11633` on `BEE_PEER_API_URL`

### DIFF
--- a/test/bee-class.browser.spec.ts
+++ b/test/bee-class.browser.spec.ts
@@ -1,5 +1,6 @@
 import { join } from 'path'
 import { beeUrl, commonMatchers } from './utils'
+import '../src'
 
 commonMatchers()
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,8 @@ jobs:
   node-tests:
     env:
       REPLICA: 5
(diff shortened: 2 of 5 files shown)


test/bee-class.browser.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { join } from 'path'
 import { beeUrl, commonMatchers } from './utils'
+import '../src'
```
vojtechsimetka, line 3: Is this committed on error?

vojtechsimetka (approved):

test/bee-class.browser.spec.ts
nugaon: oh, actually I wanted to do it in an other branch but it's a fix for test to import `window` declaration

PR #125, opened 2021-02-08
feat: feeds endpoint

agazso:
This PR implements the `/feeds` endpoints. See https://docs.ethswarm.org/api/#tag/Feed

However since at the moment there is no endpoint to update a feed, the functionality (and the tests) are quite limited. There is another PR coming with higher level feed functionality. It's more complex because it uses SOCs for uploading the correct feed updates.


src/chunk/upload.ts
```diff
@@ -0,0 +1,27 @@
...
+  chunk: SingleOwnerChunk,
+  options?: UploadOptions,
+): Promise<ReferenceResponse> {
+  const owner = bytesToHex(chunk.owner())
+  const identifier = bytesToHex(chunk.identifier())
+  const signature = bytesToHex(chunk.signature())
+  const data = serializeBytes(chunk.span(), chunk.payload())
+
+  return socAPI.upload(url, owner, identifier, signature, data, options)
+}
```
nugaon, line 27: same function defined in `src/chunk/soc.ts` (and should remain there)

src/modules/feed.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @param url         Bee URL
+ * @param owner       Owner's ethereum address in hex
+ * @param topic       Topic in hex
+ * @param options     Additional options, like type (default: 'sequence')
+ */
+export async function createFeedManifest(
+  url: string,
+  owner: string,
+  topic: string,
```
nugaon, line 23: we should provide a specific type for that with a helper function which provide this type from the given string and actually checks the given topic string satisfies the requirements for a hex topic.
this helper function should be exposed through `Utils`.

src/modules/feed.ts
```diff
@@ -0,0 +1,79 @@
...
+  at?: number
+  type?: FeedType
+}
+
+interface FeedUpdateHeaders {
+  feedIndex: string
+  feedIndexNext: string
+}
+
+export interface FindFeedUpdateResponse extends ReferenceResponse, FeedUpdateHeaders {}
```
nugaon, line 46: would be nice if the interfaces in a file preceed the functions

src/utils/bytes.ts
```diff
@@ -115,3 +115,15 @@ export function flexBytesAtOffset<Min extends number, Max extends number>(
...
+  return new Uint8Array(length) as Bytes<Length>
+}
+
+export function verifyBytesAtOffset<Length extends number>(
+  offset: number,
+  length: Length,
+  data: Uint8Array,
+): Bytes<Length> {
+  return verifyBytes(length, bytesAtOffset(offset, length, data))
+}
```
vojtechsimetka, line 129: Could you document these functions as well? Since the `utils.bytes` is now exported to the users.

src/modules/feed.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+}
+
+/**
+ * Find feed update
+ *
+ * @param url         Bee URL
+ * @param owner       Owner's ethereum address in hex
+ * @param topic       Topic in hex
+ * @param options     Additional options, like index, at, type
```
nugaon, line 61: default `type` of feed is not defined

src/modules/feed.ts
```diff
@@ -0,0 +1,79 @@
...
+
+function readFeedUpdateHeaders(headers: Dictionary<string>): FeedUpdateHeaders {
+  return {
+    feedIndex: headers['swarm-feed-index'],
+    feedIndexNext: headers['swarm-feed-index-next'],
+  }
+}
+
+/**
+ * Find feed update
```
vojtechsimetka, line 56:
```suggestion
 * Retrieve specified feed update
 * if no options are provided, retrieve the latest feed update
```

nugaon (changes requested):

src/modules/feed.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @param url         Bee URL
+ * @param owner       Owner's ethereum address in hex
+ * @param topic       Topic in hex
+ * @param options     Additional options, like index, at, type
+ */
+export async function findFeedUpdate(
```
vojtechsimetka, line 63: Probably better naming would be `retrieveFeedUpdate`

nugaon: or maybe `fetchFeed`?

vojtechsimetka:
About the tests, maybe we can just post a hardcoded value in the tests for the feeds? So really just string or array of bytes representing the feed updates and then retrieving it.

src/modules/feed.ts
agazso: This is a good idea, however I think it's outside the scope of this PR. Created an issue for it: https://github.com/ethersphere/bee-js/issues/137

src/modules/feed.ts
agazso: I usually follow the API conventions, so for example in this case when the `type` is an optional query parameter then it is also an optional parameter of the endpoint wrapper function. The default in this case is defined on the API endpoint itself.

https://docs.ethswarm.org/api/#tag/Feed/paths/~1feeds~1{owner}~1{topic}/get

src/modules/feed.ts
agazso: I usually follow the naming of the Bee documentation so that it's clear which endpoint the function refers to. However in this case I agree that this is not the best name. The issue may be that the endpoint actually does several things:
- find a feed update with the given parameters
- return the data of the feed update
- also return the index and the subsequent index of the the feed update in headers

https://docs.ethswarm.org/api/#tag/Feed/paths/~1feeds~1{owner}~1{topic}/get

src/modules/feed.ts
agazso: See my other comment explaining the weirdness around this name. With other endpoints (and the Bee interface as well) we are using `download...` and `upload...` so for the sake of consistency I think a better name would be `downloadFeedUpdate`.

WDYT?

src/modules/feed.ts
vojtechsimetka: Right, then lets just be more explicit about it. `Find feed update` is extremely cryptic.

src/modules/feed.ts
vojtechsimetka: `downdloadFeedUpdate` in me indicates downloading the actual file behind the feed. I like `fetchFeed`.

vojtechsimetka (approved):

src/modules/feed.ts
agazso: I added a comment explaining what it does.

src/modules/feed.ts
agazso: I ended up renaming it to `fetchFeedUpdate` along with a comment explaining why.

nugaon (approved):
LGTM!
feed topic helper function can be provided in the next subsequent PR

PR #134, opened 2021-02-09
feat: axios options

### LINKED ISSUE #133: Add option to override the axios configuration
It would be useful to have an extra param, for the upload/download of bytes, used to override some axios config. 

That would allow us, for instance, to use `onUploadProgress` and `cancelToken` which are useful when uploading/downloading large files to both check the progress or cancel the task.

nugaon:
On uploading and downloading big files would be nice to provide `onProgress` to keep track with the uploading/downloading and configurate other request related options.
`bee-js` make HTTP requests by `axios`, so this PR simply exposes its options on related functions and methods. 

Resolves #133

### DIFF
--- a/src/modules/file.ts
+++ b/src/modules/file.ts
@@ -1,3 +1,4 @@
+import { AxiosRequestConfig } from 'axios'
 import type { Readable } from 'stream'
 import { FileData, FileUploadOptions, UploadHeaders } from '../types'
 import { prepareData } from '../utils/data'
@@ -44,6 +45,7 @@ export async function upload(
     },
     responseType: 'json',
     params: { name },
+    ...options?.axiosOptions,
   })
 
   return response.data.reference
@@ -55,10 +57,15 @@ export async function upload(
  * @param url  Bee URL
  * @param hash Bee file hash
  */
-export async function download(url: string, hash: string): Promise<FileData<Uint8Array>> {
+export async function download(
+  url: string,
+  hash: string,
+  axiosOptions?: AxiosRequestConfig,
+): Promise<FileData<Uint8Array>> {
   const response = await safeAxios<ArrayBuffer>({
     responseType: 'arraybuffer',
     url: `${url}${endpoint}/${hash}`,
+    ...axiosOptions,
   })
   const file = {
     ...readFileHeaders(response.headers),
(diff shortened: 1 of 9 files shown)


nugaon:
Although axios' `onUploadProgress` [won't work in Node context](https://stackoverflow.com/a/55330472/10881469), but would be nice if someone could try out in browser context.

vojtechsimetka (approved):
LGTM

If you don't mind I would love to also have at least one test with some of the options like canceling request or the `onProgress` method. Just to make sure it works.

agazso:
I wouldn't expose all axios options because of two reasons:
- This way it becomes possible to override properties in a way that we cannot foresee and would break the APIs in strange ways. For example it's possible override the `url` or the `method`. This itself could be fixed by providing the `axiosOption` first, then our parameters. Still this opens up a huge surface for problems.
- It would forever tie the `bee-js` to axios (or something that is compatible with it).

Because of this I would only expose those options that are required for certain features, in this case `onUploadProgress` and `cancelToken`.

agazso (changes requested):
Please provide tests for the functionality.

nugaon:
> I wouldn't expose all axios options because of two reasons:
> 
> * This way it becomes possible to override properties in a way that we cannot foresee and would break the APIs in strange ways. For example it's possible override the `url` or the `method`. This itself could be fixed by providing the `axiosOption` first, then our parameters. Still this opens up a huge surface for problems.

I also was thinking about it, still I guess intentionally nobody would ruin basic bee funtionalities with wrong axiosOptions.
Anyway I'm still thinking about it and indeed it opens too much scenarios to use this API, that can bring also security concerns. Eventually because of the latter it should move axiosOptions to the front of the spread operation as you said.
> * It would forever tie the `bee-js` to axios (or something that is compatible with it).
with naming `axiosOptions` we can safely take out this option if somehow we would like to change this library to something else.
> 
> Because of this I would only expose those options that are required for certain features, in this case `onUploadProgress` and `cancelToken`.

Other options also could come into the picture that users need, e.g. own proxy between client app and Bee client or else. I wouldn't update gradually this feature.

vojtechsimetka:
> I wouldn't expose all axios options because of two reasons

I can get behind the reasons as well but I agree with @nugaon . Exposing and maintaining just some functions will almost always be hit/miss and probably unnecessary work/code. I also don't see a problem with locking ourselves with axios. If we need to switch (and right now I can't imagine what would be realistic reason), we just introduce breaking change in the API and ideally provide migration information.

src/modules/file.ts
```diff
@@ -55,8 +57,13 @@ export async function upload(
...
  * @param hash Bee file hash
  */
-export async function download(url: string, hash: string): Promise<FileData<Uint8Array>> {
+export async function download(
+  url: string,
+  hash: string,
+  axiosOptions?: AxiosRequestConfig,
+): Promise<FileData<Uint8Array>> {
   const response = await safeAxios<ArrayBuffer>({
+    ...axiosOptions,
```
agazso, line 66: I think in those cases where we previously didn't specify the `method` (because it defaults to `get`) from now on it would make sense to do it, otherwise it will be possible to override it with `axiosOptions`.

See also `downloadReadable` for the same.

agazso (changes requested):

src/modules/file.ts
nugaon: I don't know why anyone would change `method` to something else through `axiosOptions`.

agazso (approved):

PR #136, opened 2021-02-10
feat: feed interface

agazso:
This is a preliminary PR so that we could review and discuss the public API design together. Because of this I would ask you to focus only on the new functionality on the `Bee` interface (`makeFeedReader` and `makeFeedWriter`) and nothing else for now.

The idea here is similar to #80 but I also took the feedback from there into consideration. You can either create a feed reader or a feed writer and the writer also implements the reader interface. 

For both you have to provide the type of feed (`sequence` or `epoch`), the topic and the owner or the signer in the case of the writer. I made this the last argument so that if we want to provide a default signer with the Bee instance in the constructor that can be used as the default value.

Then you use `download` on the reader to get the latest update. `upload` on the writer (along with a reference) is used to add a new update to the feed. One surprising here is that there is a `createManifest` is on the reader because it doesn't need to be signed, unlike what the name suggests.

You can see an example how it is used in this test:
https://github.com/ethersphere/bee-js/blob/feat/feed-interface/test/bee-class.spec.ts#L217-L239

I would like ask your feedback specifically about the following questions (but feel free to give feedback about anything):
1. naming in general. I'm thinking of keeping `FeedReader` as is (or `ReadonlyFeed`) but changing `FeedWriter` to simply `Feed`
2. I feel that `type` is a bit out of the place because it's just an optional parameter on the endpoint, however if you specify it incorrectly then all your requests will be wrong. This indicates that it is not just an optional parameter but an important property of the feed and that's why I made it mandatory.
3. I started using input verification because I saw Dan using the library from Javascript. The convention I'm using is that the functions called `verify...` verifies the input provided and returns the correct type or throws exceptions if the input is incorrect. I found this easy to use with, both from the code side and both as a user. For example if you pass in an invalid topic then it will produce this output:
```
> bee.makeFeedReader('sequence', '', '')
Uncaught Error: verifyHex: not valid hex string: 
    at verifyHex (/Users/attila/Projects/bee-js/dist/index.js:19882:9)
    at verifyTopic (/Users/attila/Projects/bee-js/dist/index.js:18029:41)
    at Bee.makeFeedReader (/Users/attila/Projects/bee-js/dist/index.js:17196:50)
```
4. There is the also topic of the topics 😄 Bee expects the topic to be 64 hex characters, but it would be nice to provide a helper function so that the users can use (and compose) their own topics from strings. I already opened an issue about that https://github.com/ethersphere/bee-js/issues/137

Please use these points when replying to make the discussion easier to follow.


src/bee.ts
```diff
@@ -345,4 +351,42 @@
...
+
+  /**
+   * Make a new feed reader for downloading feed updates
+   *
+   * @param type    The type of the feed, can be 'epoch' or 'sequence'
+   * @param topic   Topic in hex or bytes
+   * @param owner   Owner's ethereum address in hex or bytes
+   */
+  makeFeedReader(
+    type: FeedType,
```
nugaon, line 363: Maybe would be better to make it optional here as well, as this is also optional in `fetchFeedUpdate`

src/bee.ts
```diff
@@ -345,4 +351,42 @@ export class Bee {
...
+    topic: Topic | Uint8Array | string,
+    owner: EthAddress | Uint8Array | string,
+  ): FeedReader {
+    const verifiedType = verifyFeedType(type)
+    const verifiedTopic = verifyTopic(topic)
+    const verifiedOwner = verifyOwner(owner)
```
nugaon, line 369: I don't get the hexToBytes transition from hex string in here, because below `makeFeedReader` transform back this from byte to hex string.

src/feed/index.ts
```diff
@@ -0,0 +1,235 @@
...
+  }
+}
+
+export async function updateFeed(
+  url: string,
+  signer: Signer,
+  topic: Topic,
+  reference: ChunkReference,
+  type: FeedType = 'sequence',
+  options?: FeedUploadOptions,
```
nugaon, line 150: type in here got default value, but this handling is not simmetric with `fetchFeedUpdate`.
I would put this type into the `options` object as well with a JSDoc, which describes what is the default value when it is undefined.

nugaon (commented):
Regards to the exposed API functions of Bee class

vojtechsimetka:
To me the interface looks good, I'd like to react to your points:

> 1. naming in general. I'm thinking of keeping `FeedReader` as is (or `ReadonlyFeed`) but changing `FeedWriter` to simply `Feed`

I am fine either way as long as this is unified with other such naming in future.

> 2. I feel that `type` is a bit out of the place because it's just an optional parameter on the endpoint, however if you specify it incorrectly then all your requests will be wrong. This indicates that it is not just an optional parameter but an important property of the feed and that's why I made it mandatory.

Yes I agree, let's make it mandatory

> 3. I started using input verification because I saw Dan using the library from Javascript. The convention I'm using is that the functions called `verify...` verifies the input provided and returns the correct type or throws exceptions if the input is incorrect. I found this easy to use with, both from the code side and both as a user. For example if you pass in an invalid topic then it will produce this output:
> 
> ```
> > bee.makeFeedReader('sequence', '', '')
> Uncaught Error: verifyHex: not valid hex string: 
>     at verifyHex (/Users/attila/Projects/bee-js/dist/index.js:19882:9)
>     at verifyTopic (/Users/attila/Projects/bee-js/dist/index.js:18029:41)
>     at Bee.makeFeedReader (/Users/attila/Projects/bee-js/dist/index.js:17196:50)
> ```

Runtime validation is off course needed and in general we should be much more mindful about it. However, this schema of returning something from a validation function is quite strange to me. Why can't it just run check? So if I take for example the `verifyTopic ` function (which really should be called `validateTopic`), it really just does typechecking and then it returns the input (which already is of a correct type BTW, the `type: unknonwn` might as well be `type: FeedType`). Then of course we have the `verifyTopic` which for some types does modify the input. I find it quite confusing, IMHO it would be much more readable if there were two functions; one to `validate` the input (which either throws an exception or just finishes with no return value) and second function to `transform` the input and return the appropriate type.

> 4. There is the also topic of the topics 😄 Bee expects the topic to be 64 hex characters, but it would be nice to provide a helper function so that the users can use (and compose) their own topics from strings. I already opened an issue about that #137

I can probably help here, adding pretty much that for PSS soon

agazso:
> Runtime validation is of course needed and in general we should be much more mindful about it. However, this schema of returning something from a validation function is quite strange to me. Why can't it just run check? So if I take for example the `verifyTopic ` function (which really should be called `validateTopic`), it really just does typechecking and then it returns the input (which already is of a correct type BTW, the `type: unknonwn` might as well be `type: FeedType`). Then of course we have the `verifyTopic` which for some types does modify the input. I find it quite confusing, IMHO it would be much more readable if there were two functions; one to `validate` the input (which either throws an exception or just finishes with no return value) and second function to `transform` the input and return the appropriate type.

Keep in mind that this is the exposed Javascript interface and that means that the type information are only used by those who are using Typescript. However if you are using Bee-js from pure Javascript nothing really prevents you to pass in any kind of data on this interfaces. You can think of this as if all types here were `unknown`. If we don't verify data it here (for example the feed type) it may happen that the Bee node will respond Bad Request and then it's much harder to find out what the problem is. However if we verify all input here then later we can trust ourselves to the type system of Typescript. 

But I understand that the current implementation is a bit strange (also it does not deal with some edge-cases). Maybe the source of confusion is that `verify...` indeed does two things: it sometimes transforms and it also validates. However I found it more difficult to separate because often it's easier to convert all possible inputs for the output type and then just verify if the output is the correct type. 

That's why the idea here is to transform the input types to a canonical internal representation and then we need to only verify that in code that can be called with user input. Then we only need to deal with that one canonical type on our internal functions. If we did the validation first, then we would need to write the validation code for all possible inputs for a certain type, then the possible transformations. For example in case of the topic:
- we would need one validation function for string input (is string?, is hex string?, is length equals 64?) to get a validated topic string that our transform function could accept
- we would need another validation function for Uint8Input (instanceof?, length?) to get the a validated topic that our transform function could accept (and then it wouldn't do anything with it).

And it gets more complicated with the `Signer` because it also accepts a private key as hex string, verifies it and converts it to a canonical private key implementation and creates a default signer with that. But if you pass in a signer it just returns that (a check is missing there obviously). In this case what would be the result of the validation?

Maybe it's easier to understand this if we call these functions `convert...` ?

> > 1. There is the also topic of the topics 😄 Bee expects the topic to be 64 hex characters, but it would be nice to provide a helper function so that the users can use (and compose) their own topics from strings. I already opened an issue about that #137
> 
> I can probably help here, adding pretty much that for PSS soon

It's not the implementation that is interesting here, we already have one:
https://github.com/ethersphere/bee-js/blob/230e645adf5f7bf1e625da149fd620d7f62a07c0/src/feed/topic.ts#L18

But rather how we want to expose it. Do we want this to be called implicitly when the user passes in a `string` as a topic? If yes what should happen if the user passes in a valid hexstring as a topic, should we do the implicit conversion as well? Shall we keep this explicit? etc.

src/bee.ts
agazso: See my response to @vojtechsimetka where I explain this in more details, but here is the gist:
the idea here is to transform the input types to a canonical internal representation and then we need to only verify that in code that can be called with user input. Then we only need to deal with that one canonical type on our internal functions.

Since we accept multiple types here it's possible that we convert from hex to bytes and then back when the caller provides hex string as input. However if they provide other types we convert to bytes and we only deal with bytes internally.

src/bee.ts
agazso: As I wrote the reasons in point 2. I'm not sure I would handle this as optional even though it's optional for the endpoint. If we want to keep this optional then I think it's better to put this as an option for `download` and `upload`.

vojtechsimetka:
> ... However if you are using Bee-js from pure Javascript nothing really prevents you to pass in any kind of data on this interfaces. You can think of this as if all types here were `unknown`. If we don't verify data it here (for example the feed type) it may happen that the Bee node will respond Bad Request and then it's much harder to find out what the problem is. However if we verify all input here then later we can trust ourselves to the type system of Typescript.

yes that is what I am saying as well, I think we maybe just don't understand each other. So let me try and demonstrate the idea. Lets take as an example the feed reader:

```ts
makeFeedReader(
  type: FeedType,
  topic: Topic | Uint8Array | string,
  owner: EthAddress | Uint8Array | string,
): FeedReader {
```
First we just validate all the types are correct.  It's great to do this on the interface rather than inside the internal functions so I like your approach.
```ts
  validateFeedType(type)
  validateTopic(topic)
  validateOwner(owner)
```

At this point we are certain, that even when used in JavaScript and some stupid type is passed, the types we actually have are correct. If not it would throw already.
  
Therefore here we can do any transformation we might want, lets say we need to transform the topic
```ts
  const topicAsBytes = transformTopic(topic)
  const ownerAsBytes = transformOwner(owner)
```
Now we just pass whatever we need to the internal function

```ts
  return makeFeedReader(this.url, type, topicAsBytes, ownerAsBytes)
}
```

> Maybe it's easier to understand this if we call these functions `convert...` ?

Yes, but then it does not really show the validation I guess. I'm just trying to be very explicit here so that the code is clear for anyone reading it.
 
> It's not the implementation that is interesting here, ... but rather how we want to expose it. Do we want this to be called implicitly when the user passes in a `string` as a topic? If yes what should happen if the user passes in a valid hexstring as a topic, should we do the implicit conversion as well? Shall we keep this explicit? etc.

Here I would advocate for no type overlapping. What I mean by that. Is a topic name `123456789abcdf...12345` of lenth 64 a hex or a string? Therefore, even if this is not user friendly, I think we need to just limit this to either `string | Uint8Array` or `hex | Uint8Array`. I lean towards the second option.

agazso:
Closes #137

nugaon (changes requested):

src/bee.ts
```diff
@@ -345,4 +351,52 @@ export class Bee {
...
+  makeFeedReader(
+    type: FeedType,
+    topic: Topic | Uint8Array | string,
+    owner: EthAddress | Uint8Array | string,
+  ): FeedReader {
+    assertIsFeedType(type)
+
+    const canonicalTopic = makeTopic(topic)
+    const canonicalOwner = makeOwner(owner)
+
```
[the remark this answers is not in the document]
agazso: We now have `assertIsFeedType` which has the return type as `asserts type is FeedType`. That is called [assertion function](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) and it is a feature of Typescript since 3.7.

We check if the value of `type` is correct here but doesn't change the value (unlike with `makeTopic` and `makeOwner`) that's why I didn't add a `makeType` function.

src/feed/type.ts
```diff
@@ -0,0 +1,12 @@
+export type FeedType = 'sequence' | 'epoch'
+
+const feedTypes = ['sequence', 'epoch']
```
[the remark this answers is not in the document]
agazso: It's better in that there is no repetition but I think it made the code a bit harder to understand, especially that I also had to change the body of `isFeedType` from:

```typescript
  return typeof type === 'string' && feedTypes.includes(type)
```
to
```typescript
  return typeof type === 'string' && feedTypes.includes(type as FeedType)
```

src/feed/index.ts
```diff
@@ -0,0 +1,232 @@
...
+  index: Index,
+  reference: ChunkReference,
+  options?: FeedUploadOptions,
+): Promise<ReferenceResponse> {
+  const identifier = makeFeedIdentifier(topic, index)
+  const at = options?.at ?? Date.now() / 1000.0
+  const timestamp = writeUint64BigEndian(at)
+  const payloadBytes = serializeBytes(timestamp, reference)
+  const cac = makeContentAddressedChunk(payloadBytes)
+  const soc = await makeSingleOwnerChunk(cac, identifier, signer)
```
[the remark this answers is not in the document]
agazso: As we discussed on the standup this is defined in the Go implementation.

BTW this is as flexible as you can get because you can always point your reference to any other chunks where you can store whatever structure you want.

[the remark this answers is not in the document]
nugaon: > BTW this is as flexible as you can get because you can always point your reference to any other chunks where you can store whatever structure you want.

The problem is not with you can't hack it, rather it could be doing in a better way.

> I think it is completely a waste of available chunk payload (in case of passing unencrypted reference, it uses 40 bytes from <4096 bytes).

also it could refer more than one chunk for exciting other use-cases

> Secondly, what if I want to add to this update more than one reference of swarm content? Or I want to attach metadata as well? This rigid data structure does not let me do that.

src/modules/feed.ts
```diff
@@ -1,23 +1,29 @@
...
-
 export interface CreateFeedOptions {
   type?: FeedType
 }
-export interface FindFeedUpdateOptions {
-  index?: number
+
+export interface FeedUpdateOptions {
+  /**
+   * Specifies the start date as unix time stamp
```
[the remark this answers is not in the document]
agazso: In a feed update there is a `timestamp` and a `reference` (See https://github.com/ethersphere/bee-js/blob/e591ce3c5643c1c78181b8df59235b37e87a8f3a/src/feed/index.ts#L37). If you specify the `at` then the lookup will find the first feed update that has timestamp greater than the `at` value.

You can find the documentation of the actual [Find feed update](https://docs.ethswarm.org/api/#tag/Feed/paths/~1feeds~1{owner}~1{topic}/get) endpoint. Please note that even though `index` is in the documentation the actual implementation does not use it, so currently you cannot lookup based on the index.

test/bee-class.spec.ts
```diff
@@ -199,4 +202,49 @@ describe('Bee class', () => {
...
+
+      expect(firstUpdateReferenceResponse.reference).toEqual(bytesToHex(referenceZero))
+      expect(firstUpdateReferenceResponse.feedIndex).toEqual('0000000000000000')
+
+      const referenceOne = new Uint8Array([...new Uint8Array([1]), ...new Uint8Array(31)]) as ChunkReference
+
+      // TODO without this the test fails quite often
+      // with the sleep it's better but still fails sometimes
+      // there may be a race condition during lookup
+      await sleep(1 * 1000)
```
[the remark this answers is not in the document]
agazso: It can be removed for now because sequence based feeds currently does not use the timestamp for lookup.

test/feed/index.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+describe('feed', () => {
+  const url = beeUrl()
+  const owner = stripHexPrefix(testIdentity.address)
+  const signer = makeDefaultSigner(hexToBytes(testIdentity.privateKey) as PrivateKey)
+  const topic = '0000000000000000000000000000000000000000000000000000000000000000' as HexString
+
+  test('empty feed update', async () => {
+    const emptyTopic = '1000000000000000000000000000000000000000000000000000000000000000' as HexString
```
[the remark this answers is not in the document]
agazso: If you look at the code you won't find many type assertions. We only need the `as` type assertions in tests where we don't always check the validity of the types and only for testing internal functionality which already uses strict types. 

For external functionality (that is what the users will use) we are using plain Javascript types (like `string` and `Uint8Array`) and convert them to strict types and pass them to internal functions, so the users should not use type assertions.

It would be also possible to avoid the type assertions in the tests by using the verified constructors of the types (in this example `verifyHex`). It's also possible to change to `FlavoredType` as you mention however from that point this becomes only a type annotation and the compiler won't enforce the right types.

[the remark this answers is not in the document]
nugaon: > It's also possible to change to FlavoredType as you mention however from that point this becomes only a type annotation and the compiler won't enforce the right types.

It actually enforces between of flavor types, that we need.
I just don't see anywhere we would do this branded type checks, but actually if we would do, then all tests went wrong as you said because those are using these type assertions (to mimic a FlavoredType use-case).

I checked and this type assertions are also in other place in the source code that could yield the before mentioned problem.

https://github.com/ethersphere/bee-js/blob/8e95aafb803331061a42cf388b9dc8952945a699/src/chunk/cac.ts#L43

src/feed/index.ts
```diff
@@ -0,0 +1,232 @@
...
+    throw e
+  }
+}
+
+export async function updateFeed(
+  url: string,
+  signer: Signer,
+  topic: Topic,
+  reference: ChunkReference,
+  options: FeedUploadOptions,
```
nugaon, line 151: why is it not optional?

src/feed/index.ts
```diff
@@ -0,0 +1,232 @@
...
+  upload(reference: ChunkReference | Reference, options?: FeedUploadOptions): Promise<ReferenceResponse>
+}
+
+export function isEpoch(epoch: unknown): epoch is Epoch {
+  return typeof epoch === 'object' && epoch !== null && 'time' in epoch && 'level' in epoch
+}
+
+function hashFeedIdentifier(topic: Topic, index: IndexBytes): Identifier {
+  return keccak256Hash(topic, index)
+}
```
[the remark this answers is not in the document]
agazso: It provides two benefits:
- documentation that the feed identifier is made of the hash of the topic and index
- and type safety because it expects topic (which is exactly 32 bytes) and an index (which is exactly 8 bytes) as arguments. If you pass in different arguments it will be a compile error which is not the case with calling `keccak256hash` directly.

From my experience many hours can be wasted to verify if you hash the correct things together. This (and strong typing in general) helps avoiding mistakes.

src/feed/index.ts
agazso: fixed

test/feed/index.spec.ts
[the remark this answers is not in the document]
agazso: I think this is an important topic to discuss but it's not the scope of this (already big) PR. I created a separate issue for that
https://github.com/ethersphere/bee-js/issues/143

src/feed/index.ts
```diff
@@ -0,0 +1,232 @@
...
+  readonly owner: Owner
+  readonly topic: Topic
+  /**
+   * Download the latest feed update
+   */
+  download(options?: FeedUpdateOptions): Promise<FetchFeedUpdateResponse>
+  /**
+   * Create feed manifest chunk and return the reference
+   */
+  createManifest(): Promise<ReferenceResponse>
```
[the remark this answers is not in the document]
agazso: Yeah, on a second thought it may be better to put it on the Bee class.  I just changed it.

https://github.com/ethersphere/bee-js/pull/136/commits/9b2c46b3385fba7652851328997c985e612828b6

nugaon (approved):
LGTM, although I hope we will have generic interface later on feed payload.

vojtechsimetka:
What do we do about the tests? Should we disable them now on the CI until it is fixed in bee? This is important for other PRs who will have some tests randomly failing.

AuHau:
To my understanding we are waiting on Bee release that should fix that.

agazso:
> To my understanding we are waiting on Bee release that should fix that.

Exactly, the new release with fixed feeds are scheduled for today. Then I will push a commit to this branch to test with the new Bee version and only merge to master if all tests work fine.

src/bee.ts
```diff
@@ -345,4 +353,72 @@ export class Bee {
...
+   * Create feed manifest chunk and return the reference to it
+   *
+   * @param type    The type of the feed, can be 'epoch' or 'sequence'
+   * @param topic   Topic in hex or bytes
+   * @param owner   Owner's ethereum address in hex or bytes
+   */
+  createFeedManifest(
+    type: FeedType,
+    topic: Topic | Uint8Array | string,
+    owner: EthAddress | Uint8Array | string,
```
AuHau, line 367: Shouldn't this be `OwnerInput` as presented in `/src/chunk/owner.ts`?

src/feed/index.ts
```diff
@@ -0,0 +1,226 @@
...
+  timestamp: number
+  reference: ChunkReference
+}
+
+/**
+ * FeedReader is an interface for downloading feed updates
+ */
+export interface FeedReader {
+  readonly type: FeedType
+  readonly owner: Owner
```
AuHau, line 47: Honestly, I think the type and the property should be named `OwnerAddress` as `Owner` IMO does not really explain what it actually contains. Eq. it could be some object with metadata etc...

agazso: Good point, created an issue for that
https://github.com/ethersphere/bee-js/issues/152

src/bee.ts
agazso: I think this better conveys what the expected type on the public interface, because it doesn't require users to follow up on what `OwnerInput` is. In other words `OwnerInput` is more like an internal type and this is our external interface. And fortunately the compiler verifies that it's actually correct :)

AuHau: Yeah, make sense! I will use the same approach in the SOC PR. 
Although I am thinking if then it make sense to have the `OwnerInput` type? I mean it is a sort of duplication and one might run into a problem when wants to update `OwnerInput` type and might expect that the change will propagate everywhere where OwnerAddress is used 🤔 Not sure if I am not overthinking this one though 😅

agazso: I had this idea to separate the input types from output types but I'm not sure anymore that this is actually useful. Feel free to remove `OwnerInput`.

PR #139, opened 2021-02-17
fix: remove automatic content-type header on file upload

### LINKED ISSUE #138: uploadFile sets the content-type to application/octet-stream
Because of this all uploaded files' content-type automatically become `application/octet-stream`. However if there is no `content-type` header specified Bee tries to detect the content type.

I think it would be better to remove the automatic `content-type` header and either let Bee detect it or let the user set it explicitly.

agazso:
Fixes #138

### DIFF
--- a/src/modules/file.ts
+++ b/src/modules/file.ts
@@ -42,7 +42,6 @@ export async function upload(
     url: url + endpoint,
     data: prepareData(data),
     headers: {
-      'content-type': 'application/octet-stream',
       ...extractFileUploadHeaders(options),
     },
     responseType: 'json',


nugaon (approved):
LGTM!

nugaon:
Although, we could take out `contentType` from `FileUploadOptions` as well then, am I right?
Also it has the similar assignment in chunk.ts, and soc.ts

agazso:
> Although, we could take out `contentType` from `FileUploadOptions` as well then, am I right?

No, we do want to let the user define the content-type. What this change does is that when the user does not define a content-type we don't set it in the header automatically to `application/octet-stream` but instead let Bee to decide it based on the content of the file.

> Also it has the similar assignment in chunk.ts, and soc.ts

It only matters with files, because there the content-type is preserved and returned when the user fetches the content. Chunks (SOCs included) are handled as binaries by Bee and therefore having content-type as `application/octet-stream` is correct.

vojtechsimetka (approved):

PR #145, opened 2021-02-24
fix: release 0.5.1

agazso:
There is a regression in Bee 0.5.1 that it no longer allows to upload empty files in collections. This is a workaround to put something into the empty files, so that upload won't fail.

https://github.com/ethersphere/bee/issues/1332

However empty files (and directories) were deliberately chosen to be part of the test data so this should be reverted once this issue is fixed in Bee.


vojtechsimetka (approved):
Will you please create issue to revert the changes?

agazso:
> Will you please create issue to revert the changes?

Good point, I opened an issue

https://github.com/ethersphere/bee-js/issues/146

PR #147, opened 2021-02-24
feat: exposing soc interface

### LINKED ISSUE #71: Exposing SOC and Feed APIs on the Bee class

AuHau:
Initial draft PR of exposing Single Owner Chunk API on `Bee` class, that will superceed #80 PR. So far without tests, those will be added tomorrow.

The PR builds on top of `feat/feed-interface` branch, in order not to mixup the changes from that branch and once it will be merged I will rebase it to `master`. 

Closes #71


src/chunk/soc.ts
```diff
@@ -137,3 +164,36 @@ export function uploadSingleOwnerChunk(
...
+/**
+ * Helper function to create and upload SOC.
+ *
+ * @param url         The url of the Bee service
+ * @param signer      The singer interface for signing the chunk
+ * @param identifier  The identifier of the chunk
+ * @param data        The chunk data
+ * @param options
+ */
+export async function makeAndUploadSingleOwnerChunk(
```
agazso, line 177: The convention in the existing code is that `makeX` functions return type X (similar to a constructor). This is not the case here, so maybe it would be better to call this `uploadSingleOwnerChunkData`.

AuHau: Ok, good idea! I a bit struggled how to name it, so I just combined those two.

src/feed/index.ts
```diff
@@ -2,7 +2,7 @@ import { makeContentAddressedChunk } from '../chunk/cac'
 import { keccak256Hash } from '../chunk/hash'
 import { serializeBytes } from '../chunk/serialize'
 import { EthAddress, Signer } from '../chunk/signer'
-import { Identifier, makeSingleOwnerChunk, verifySingleOwnerChunk } from '../chunk/soc'
+import { Identifier, makeSingleOwnerChunk, uploadSingleOwnerChunkData, verifySingleOwnerChunk } from '../chunk/soc'
```
agazso, line 5: Please address the warnings before merging.

It's possible to run the checks locally by running `npm run lint:check`. Running `npm run lint` also fixes the formatting for you.

agazso (approved):
Looks good to me!

Please keep in mind that this PR is also needs to be merged to `feat/feed-interface` so depending which gets merged first, there may be a conflict with this.

https://github.com/ethersphere/bee-js/pull/151

test/chunk/signer.spec.ts
```diff
@@ -22,4 +22,45 @@ describe('signer', () => {
 
     expect(bytesToHex(recoveredAddress, true)).toEqual(testIdentity.address)
   })
+
+  describe('makeSigner', () => {
+    test('converts string', async () => {
+      const signer = makeSigner(testIdentity.privateKey)
+      const signature = await sign(dataToSign, signer)
+
+      expect('0x' + bytesToHex(signer.address)).toEqual(testIdentity.address)
```
nugaon, line 31:
```suggestion
      expect(bytesToHex(signer.address, true)).toEqual(testIdentity.address)
```

test/chunk/signer.spec.ts
```diff
@@ -22,4 +22,45 @@ describe('signer', () => {
...
+      expect(signature).toEqual(expectedSignature)
+    })
+
+    test('converts uintarray', async () => {
+      const signer = makeSigner(hexToBytes(testIdentity.privateKey))
+      const signature = await sign(dataToSign, signer)
+
+      expect('0x' + bytesToHex(signer.address)).toEqual(testIdentity.address)
```
nugaon, line 39:
```suggestion
      expect(bytesToHex(signer.address, true)).toEqual(testIdentity.address)
```

nugaon (approved):
LGTM!

PR #149, opened 2021-02-24
feat: eth utils for PSS

vojtechsimetka:


src/utils/eth.ts
```diff
@@ -74,19 +78,30 @@ export function toLittleEndian(bigEndian: number | string | HexString, pad = 2):
...
  *
- * @param ethAddress
- * @param networkId
+ * @param ethAddress  Public ethereum address
+ * @param networkId   Swarm network id
+ *
+ * @return Swarm overlay address
  */
 export function ethToSwarmAddress(ethAddress: string | HexString | EthAddress, networkId = 1): OverlayAddress {
+  if (!isEthAddress(ethAddress)) throw new TypeError('incorrect ETH address')
```
AuHau, line 99: I think you could create and use `assertIsEthAddress()` function as Attila created in the Feeds PR [here](https://github.com/ethersphere/bee-js/pull/136/files#diff-64091434a876d44403f86d6b5d7a07f4bfe959324fc90d992fe1a033d2d3bc43R7).

src/utils/eth.ts
```diff
@@ -0,0 +1,110 @@
+import { keccak256, sha3_256 } from 'js-sha3'
+import { BrandedString } from '../types'
+import { HexString, hexToBytes, intToHex, isHexString, stripHexPrefix, verifyHex } from './hex'
+export type EthAddress = BrandedString<'EthAddress'>
```
agazso, line 4: We already have a type named `EthAddress` here:
https://github.com/ethersphere/bee-js/blob/d6ac8b0b52093289725f8e199bbd45fc457b38a9/src/chunk/signer.ts#L13

Maybe this could be called `HexEthAddress`?

src/utils/eth.ts
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * Pretty much typed version from web3js
+ * https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-utils/src/utils.js
+ *
+ * @param address  Ethereum address
+ *
+ * @return True if is valid eth address
+ */
+export function isEthAddress(address: string | HexString): address is EthAddress {
+  if (typeof address !== 'string' || !/^(0x)?[0-9a-f]{40}$/i.test(address)) {
```
agazso, line 18: I find this if...else construct a bit hard to follow. The first if is a negative check and returns false, then there is a positive check that returns true, then the else case can also return false under certain circumstances.

In general in these kinds of functions it's better to eliminate all the negative cases and only return true at the end. Or alternatively return true for the positive cases and return false at the end.

Maybe it's also worth considering extracting the conditions to predicate functions, that usually helps explaining what is going on.

src/utils/hex.ts
```diff
@@ -33,7 +33,7 @@ export function hexToBytes(hex: HexString): Uint8Array {
 /**
  * Converts array of number or Uint8Array to hex string.
  *
- * Optionally provides a the '0x' prefix.
+ * Optionally provides'0x' prefix.
```
agazso, line 36: Typo: missing space between `provides` and `'0x'`

src/utils/hex.ts
```diff
@@ -46,13 +46,33 @@ export function bytesToHex(bytes: Uint8Array, withPrefix = false): HexString {
   return `${prefix}${hex}` as HexString
 }
 
+/**
+ * Converst integer number to hex string.
+ *
+ * Optionally provides'0x' prefix or padding
```
agazso, line 52: Typo: missing space between provides and '0x'

src/utils/eth.ts
vojtechsimetka: My aim here was to minimally modify the reference function that is directly implemented in `web3js`. I will try and create the predicates to see if that helps.

vojtechsimetka: Ok this actually makes sense, let me know what do you think.

src/utils/eth.ts
```diff
@@ -0,0 +1,132 @@
...
+    typeof address === 'string' && (/^(0x|0X)?[0-9a-f]{40}$/.test(address) || /^(0x|0X)?[0-9A-F]{40}$/.test(address))
+  )
+}
+
+/**
+ * Check if this is checksummed ethereum address
+ *
+ * @param address Ethereum address as hex string
+ */
+function isEthAddrChecksummed(address: string | HexString | HexEthAddress): address is HexEthAddress {
```
AuHau, line 24: The name of this function sort of implies that it checks only if the address has checksum on it (eq. some chars lowercase and some uppercase, pretty much what `isEthAddCaseIns` function) not that it actually checks the validity of the checksum. Something like `checkEthAddrChecksum` or `hasValidEthChecksum` might apply better here.

But it might just my subjective impression 😅

AuHau (commented):
Generally LGTM, only one small nitpick. I will leave approval to others though.

src/utils/eth.ts
agazso: How about `isChecksummedEthAddress`?

src/utils/hex.ts
```diff
@@ -46,13 +46,33 @@ export function bytesToHex(bytes: Uint8Array, withPrefix = false): HexString {
...
+}
+
 /**
  * Type guard for HexStrings
  *
  * @param s string input
  */
 export function isHexString(s: string): s is HexString {
-  return /^(0x)?[0-9a-fA-F]+$/i.test(s)
+  return typeof s === 'string' && /^(0x)?[0-9a-f]+$/i.test(s)
```
agazso, line 75: This changes the functionality to only allow lowercase hex strings. Was that the intention?

I'm not for or against either way, but it would be good to indicate that this is the case, at least in the comment.

vojtechsimetka: No it doesn't. The `i` at the end indicates that the match is case insensitive. This way it will accept all of these:
```
0xab
0XAB
0xAb
```

vojtechsimetka: So the `A-F` was actually completely redundant. I also missed it first time around.

agazso: Ah ok 👍

src/utils/eth.ts
vojtechsimetka: I like `isValidChecksummedEthAddress`, will rename

agazso (approved):

PR #151, opened 2021-02-25
fix: upload SOC twice

agazso:
Add functionality to delete chunks from local storage to make tests repeatable. Also add helper functions for tests to ignore certain errors.

This PR is forked from `feat/feed-interface` not from master to make the review easier. Once it is approved it will be merged back to it, then that can be merged back to master.

### DIFF
--- a/src/modules/debug/chunk.ts
+++ b/src/modules/debug/chunk.ts
@@ -1,0 +1,39 @@
+import { BeeResponse } from '../../types'
+import { safeAxios } from '../../utils/safeAxios'
+
+const endpoint = '/chunks'
+
+/**
+ * Check if chunk at address exists locally
+ *
+ * @param url      Bee debug url
+ * @param address  Swarm address of chunk
+ *
+ * @returns BeeResponse if chunk is found or throws an exception
+ */
+export async function checkIfChunkExistsLocally(url: string, address: string): Promise<BeeResponse> {
+  const response = await safeAxios<BeeResponse>({
+    url: url + endpoint + `/${address}`,
+    responseType: 'json',
+  })
+
+  return response.data
+}
+
+/**
+ * Delete a chunk from local storage
+ *
+ * @param url      Bee debug url
+ * @param address  Swarm address of chunk
+ *
+ * @returns BeeResponse if chunk was deleted or throws an exception
+ */
+export async function deleteChunkFromLocalStorage(url: string, address: string): Promise<BeeResponse> {
+  const response = await safeAxios<BeeResponse>({
+    method: 'delete',
+    url: url + endpoint + `/${address}`,
+    responseType: 'json',
+  })
+
+  return response.data
+}
--- a/test/feed/index.spec.ts
+++ b/test/feed/index.spec.ts
@@ -1,11 +1,13 @@
 import { fetchFeedUpdate } from '../../src/modules/feed'
 import { HexString, hexToBytes, stripHexPrefix, verifyHex } from '../../src/utils/hex'
 import { beeUrl, testIdentity } from '../utils'
-import { ChunkReference, downloadFeedUpdate, findNextIndex, uploadFeedUpdate } from '../../src/feed'
+import { ChunkReference, downloadFeedUpdate, findNextIndex, Index, uploadFeedUpdate } from '../../src/feed'
 import { Bytes, verifyBytes } from '../../src/utils/bytes'
-import { makeDefaultSigner, PrivateKey } from '../../src/chunk/signer'
+import { makeDefaultSigner, PrivateKey, Signer } from '../../src/chunk/signer'
 import { makeContentAddressedChunk } from '../../src/chunk/cac'
 import * as chunkAPI from '../../src/modules/chunk'
+import { Topic } from '../../src/feed/topic'
+import { BeeResponseError } from '../../src'
 
 function makeChunk(index: number) {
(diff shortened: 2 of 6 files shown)


test/feed/index.spec.ts
```diff
@@ -18,6 +20,18 @@ async function uploadChunk(url: string, index: number): Promise<ChunkReference>
   return hexToBytes(referenceResponse.reference as HexString) as ChunkReference
 }
 
+async function tryUploadFeedUpdate(url: string, signer: Signer, topic: Topic, index: Index, reference: ChunkReference) {
```
AuHau, line 23: This seems a bit wrong to me. Prior to running the test we should set up a valid state and then we should not ignore any errors (eq. to "try"), because the outcome should be correct or there is some problem and then it should fail.

src/modules/debug/chunk.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Check if chunk at address exists locally
+ *
+ * @param url      Bee debug url
+ * @param address  Swarm address of chunk
+ *
+ * @returns BeeResponse if chunk is found or throws an exception
+ */
+export async function checkIfChunkExistsLocally(url: string, address: string): Promise<BeeResponse> {
```
AuHau, line 14: I mean, doesn't this functionality gives the `/src/modules/chunk` `download()` function? Maybe lets limit the code duplication?

test/utils.ts
```diff
@@ -114,6 +115,19 @@ export function beeDebugUrl(url: string = beeUrl()): string {
...
+/**
+ * Try to delete a chunk from local storage, ignoring all errors
+ *
+ * @param address  Swarm address of chunk
+ */
+export async function tryDeleteChunkFromLocalStorage(address: string): Promise<void> {
+  try {
+    await deleteChunkFromLocalStorage(beeDebugUrl(), address)
+  } catch (e) {
+    // ignore errors
```
AuHau, line 127: I would be a bit cautious about ignoring all errors. For example, the Bee node might have turned off Debug API and this would not catch it (although then I guess will fail other tests as well). Could it be narrowed down to ignore only "not found" errors?

src/modules/debug/chunk.ts
agazso: It's not really the same thing: chunk download will try to download a chunk from the network, whereas this function only checks if the file is available on your local node or not.

We are not using this functionality at the moment but I just added for the sake of completeness.

test/feed/index.spec.ts
agazso: This function is only supposed to be used for preparing the node state for finding feed updates and not for testing feed updates. I tried to emphasize it with the naming, maybe adding a comment would make it more clear?

src/modules/debug/chunk.ts
AuHau: Hmm I see, I am just a bit confused that it uses the same endpoint `/chunks/<address>` so Bee node won't download it because you request `responseType: json` instead of `arraybuffer`? (that is the only difference that I found)

AuHau: Oh this is Debug API we are talking about 🙈 Sorry! Make sense then :-)

test/feed/index.spec.ts
AuHau: Well, my problem is, what will happen if the update won't happen? Eq. there was already some SOC chunk on the given index, so the referenced CAC won't get updated. Then the test might fail for some assertions on what is expected to be referenced...

Generally, this approach depends on to state prior to running the test case which in my experience is a bad thing. Every test should be as isolated as possible. For example the order of tests execution might change and then suddenly there might be unexpected failures, or you run only limited set of test cases, which then might lead to same problem, or you add new test later on that might upload SOC to that address and introduce conflicts...

agazso: This workaround is only for make the tests repeatable in a local environment without the need to stop the nodes, delete the data and restart the nodes. On CI we have a new, clean environment all the time. This guarantees that separate test runs are reproducible. It's also possible to do this with the provided `/test/bee.sh` script using the `--ephemeral` switch, but it's inconvenient to do so. We also considered having a debug endpoint to delete all data from a node that could be invoked before running our tests but that's also not trivial to do because of the nature of a decentralized push synching that the Bee nodes do.

https://github.com/ethersphere/bee/issues/1264

Also please note that these tests are independent of each other because they use different topics. The function in question is actually used to help setting up a test where we are checking if the lookup works for multiple updates, so it's not that important what the uploaded content is. Of course it's possible to write data to your local test cluster that will make tests fail. In that case (or if you are in doubt regarding your tests) you can always stop the test nodes, delete all data, restart and see if the problem persist.

vojtechsimetka: I also don't like this solution long term and would love to have some `// FIXME: ` note here and probably issue about it with `need help` label.

vojtechsimetka (approved):

test/feed/index.spec.ts
agazso: Done, see https://github.com/ethersphere/bee-js/issues/154

nugaon (approved):

src/modules/debug/chunk.ts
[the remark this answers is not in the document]
agazso: I opened an issue https://github.com/ethersphere/bee-js/issues/156

PR #155, opened 2021-02-26
feat: expose the getPeers function on `BeeDebug`

### LINKED ISSUE #153: getPeers() not supported in BeeDebug?
Connectivity has a getPeers() method that is used elsewhere, but not exported?  This is the only debugAPI interface standing between me and releasing a node monitor written in node.js.

Declaration:
https://github.com/ethersphere/bee-js/blob/313830adb9d7db0c43f605b9d3db69f6ea92a07d/src/modules/debug/connectivity.ts#L27

Reference:
https://github.com/ethersphere/bee-js/blob/313830adb9d7db0c43f605b9d3db69f6ea92a07d/test/modules/pss.spec.ts#L17

But no method in BeeDebug?:
https://github.com/ethersphere/bee-js/blob/313830a/src/bee-debug.ts

vojtechsimetka:
resolves #153

### DIFF
--- a/src/bee-debug.ts
+++ b/src/bee-debug.ts
@@ -24,6 +24,13 @@ export class BeeDebug {
     return nodeAddresses.pss_public_key
   }
 
+  /**
+   * Get list of peers for this node
+   */
+  getPeers(): Promise<connectivity.Peers> {
+    return connectivity.getPeers(this.url)
+  }
+
   /*
    * Balance endpoints
    */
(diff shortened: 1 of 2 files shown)


agazso (approved):
It would be good to have at least one test checking if this endpoint really returns what we expect.

vojtechsimetka:
Makes sense, I'll do it before merge

PR #158, opened 2021-02-26
feat: expose types for BeeDebug

### LINKED ISSUE #157: Exposing BeeDebug types
We don't expose the types used in the `BeeDebug` class like `Peer, Peers, PeerBalance, BalanceResponse...`. Currently these types are not reachable by our users.

vojtechsimetka:
Resolved #157

### DIFF
--- a/src/modules/debug/connectivity.ts
+++ b/src/modules/debug/connectivity.ts
@@ -1,12 +1,5 @@
 import { safeAxios } from '../../utils/safeAxios'
-
-export interface NodeAddresses {
-  overlay: string
-  underlay: string[]
-  ethereum: string
-  public_key: string
-  pss_public_key: string
-}
+import { NodeAddresses, Peer } from '../../types'
 
 export async function getNodeAddresses(url: string): Promise<NodeAddresses> {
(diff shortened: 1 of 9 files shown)


agazso (approved):

src/modules/debug/connectivity.ts
```diff
@@ -16,19 +9,15 @@ export async function getNodeAddresses(url: string): Promise<NodeAddresses> {
...
-export async function getPeers(url: string): Promise<Peers> {
+export async function getPeers(url: string): Promise<Peer[]> {
   const response = await safeAxios<Peers>({
     url: url + '/peers',
     responseType: 'json',
   })
 
-  return response.data
+  return response.data.peers
 }
```
vojtechsimetka, line 23: Please notice I decided to change the type to directly return `Peer[]` array rather than object that would contain it. I find this much easier to use. I also noticed there are other functions that almost have the same issue but were not quite so easy to justify as this one.

agazso: I also like it better this way. I'm also happy that our tests caught this kind of change 😄

PR #180, opened 2021-03-04
fix(pss): subscribe in browsers, removed readable in browsers

vojtechsimetka:
This adds support for the PSS in browsers and tests send + receive in browser context. I've had to create a browser specific version of the `utils/data.ts` which does not use `Buffer` as it does not exist in browsers. The `prepareWebsocketData` now also handles `Blob` which seems to be returned from bee in browser context.


src/utils/data.ts
```diff
@@ -11,14 +11,22 @@ export function prepareData(data: string | ArrayBuffer | Uint8Array | Readable):
...
+export async function prepareWebsocketData(data: Data | Blob): Promise<Uint8Array> | never {
   if (typeof data === 'string') {
     return new TextEncoder().encode(data)
   } else if (data instanceof Buffer) {
     return new Uint8Array(data)
   } else if (data instanceof ArrayBuffer) {
     return new Uint8Array(data)
-  } else {
+  } else if (data instanceof Blob) {
+    return new Uint8Array(await new Response(data as Blob).arrayBuffer())
```
agazso, line 26: Why do we need a `Response` object? Isn't this enough?

https://developer.mozilla.org/en-US/docs/Web/API/Blob/arrayBuffer

vojtechsimetka: It has a worse compatibility (e.g. Opera does not support it at all) and `Response` object is recommended.

src/utils/data.ts
```diff
@@ -11,14 +11,22 @@ export function prepareData(data: string | ArrayBuffer | Uint8Array | Readable):
   return data
 }
 
-export function prepareWebsocketData(data: Data): Uint8Array {
+function isBufferArray(data: unknown) {
```
agazso, line 14: Please provide return type (I thought we had a linter rule for this 😄 )

agazso (commented):
As I understand the issue here is that this was never tested in browsers and the node.js and browser versions are different enough so that this is not working at all?

vojtechsimetka:
Exactly. The reason is that the type of the message data in browser for WebSocket can be any type [[docs](https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent/data)] (and seem like it is Blob in this case) so it was failing to convert and we don't have tests for this (and never tried).

We use the [isomorhpic-ws](https://github.com/heineiuo/isomorphic-ws) to have compatibility between browser and nodejs but it seems it's not 100% at least when it comes to types (and unclear what types are really allowed in the nodejs case)

agazso:
In that case let's write browser tests for this use case 😄

src/utils/data.browser.ts
```diff
@@ -0,0 +1,23 @@
...
+export async function prepareWebsocketData(data: string | ArrayBuffer | Blob): Promise<Uint8Array> | never {
+  if (typeof data === 'string') {
+    return new TextEncoder().encode(data)
+  } else if (data instanceof ArrayBuffer) {
+    return new Uint8Array(data)
+  } else if (data instanceof Blob) {
+    return new Uint8Array(await new Response(data as Blob).arrayBuffer())
+  }
+
+  throw new TypeError('unknown websocket data type')
```
nugaon, line 22: See comment _switch-case_

src/utils/data.browser.ts
```diff
@@ -0,0 +1,23 @@
+import type { Readable } from 'stream'
+
+export function prepareData(data: string | ArrayBuffer | Uint8Array | Readable): Uint8Array | Readable {
+  if (typeof data === 'string') {
+    return new TextEncoder().encode(data)
+  } else if (data instanceof ArrayBuffer) {
+    return new Uint8Array(data)
+  }
```
nugaon, line 8: See comment _switch-case_

jest-puppeteer.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  launch: {
+    dumpio: true,
+  },
+}
```
nugaon, line 5: can you give a context about this pupeteer.config file?

test/bee-class.browser.spec.ts
```diff
@@ -80,4 +83,76 @@ describe('Bee class - in browser', () => {
...
+        )
+
+        expect(result).toEqual(message)
+        done()
+      },
+      PSS_TIMEOUT,
+    )
+
+    it(
+      'should send and receive pss message',
```
nugaon, line 123: Please amend this title according to what is happening exactly which differs from the previous test.

test/utils.ts
```diff
@@ -146,7 +146,7 @@ export const okResponse: BeeResponse = {
   code: 200,
   message: 'OK',
 }
-export const PSS_TIMEOUT = 60000
+export const PSS_TIMEOUT = 120000
```
nugaon, line 149: noice

src/utils/data.ts
```diff
@@ -11,14 +11,22 @@ export function prepareData(data: string | ArrayBuffer | Uint8Array | Readable):
   return data
 }
 
-export function prepareWebsocketData(data: Data): Uint8Array {
+function isBufferArray(buffer: unknown): buffer is Buffer[] {
+  return Array.isArray(buffer) && buffer.length > 0 && buffer.reduce((prev, cur) => prev && cur instanceof Buffer, true)
```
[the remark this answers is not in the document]
nugaon: Though, I'm still on the opinion it is wastefull process time, and we could be optimistic if the first element is something of an Array, then the other as well in case of data objects.

src/utils/data.ts
```diff
@@ -11,14 +11,22 @@ export function prepareData(data: string | ArrayBuffer | Uint8Array | Readable):
...
   } else if (data instanceof ArrayBuffer) {
     return new Uint8Array(data)
-  } else {
+  } else if (data instanceof Blob) {
+    return new Uint8Array(await new Response(data as Blob).arrayBuffer())
+  } else if (isBufferArray(data)) {
     return new Uint8Array(Buffer.concat(data))
   }
+
+  throw new TypeError('unknown websocket data type')
```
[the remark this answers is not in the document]
vojtechsimetka: I don't think this can be turned into switch-case nicely as the conditions are not straight forward. That being said I changed it from `if-else if` to hopefully more readable form of 
```
if (...) return ...

if (...) return ...

throw new TypeError(...)
```

AuHau (approved):
LGTM

src/utils/data.ts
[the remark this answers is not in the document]
nugaon: The `switch ... case` can yield better execution time and also improve the readability.

PR #193, opened 2021-03-05
ci: separated browser and nodejs tests

vojtechsimetka:
Splits the tests into nodejs and browser job.


nugaon (commented):
Can you explain why is it important to split the tests into two separated jobs?
It just seems unnecessary code duplication for me at the first glance.

vojtechsimetka:
> Can you explain why is it important to split the tests into two separated jobs?
> It just seems unnecessary code duplication for me at the first glance.

We were discussing with @agazso that it makes sense to separate the nodejs and browser tests and that the browser ones don't need to be run on each nodejs version. I know and agree there is a code duplication, but I don't know how else to do it. Any suggestions are welcome.

AuHau (approved):
LGTM, I was also searching for a way how to run "background job/container" but did not really find a way, so I guess for now we have to duplicate.

AuHau:
Btw. you will have to remove the required checks in order to merge it ;-)

PR #192, opened 2021-03-05
ci(release-please): before 1.x any breaking changes will bump minor

vojtechsimetka:
…version instead of major

PR #198, opened 2021-03-10
test(feed): increase flaky feed tests timeout

vojtechsimetka:
As requested in https://github.com/ethersphere/bee-js/pull/192#issuecomment-792734775

### DIFF
--- a/test/utils.ts
+++ b/test/utils.ts
@@ -147,6 +147,7 @@ export const okResponse: BeeResponse = {
   message: 'OK',
 }
 export const PSS_TIMEOUT = 60000
+export const FEED_TIMEOUT = 120000
 
 export const testChunkPayload = new Uint8Array([1, 2, 3])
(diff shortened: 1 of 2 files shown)


test/utils.ts
```diff
@@ -147,6 +147,7 @@ export const okResponse: BeeResponse = {
   message: 'OK',
 }
 export const PSS_TIMEOUT = 60000
+export const FEED_TIMEOUT = 120000
```
AuHau, line 150: Do you think it is possible to generalize it like this? I would expect that maybe some tests will require different values.

AuHau (approved):
LGTM

test/utils.ts
vojtechsimetka: Good point. I was just following the PSS example TBH. But I do like having some constant like this and then potentially in the tests themselves doing multiplications.

PR #201, opened 2021-03-10
feat: hex string type length and validation

AuHau:
So this PR is very much WIP and there is currently quite some errors in it so please look more high-level at the approach rather on the details. Looking at the code I feel there might be a bit bigger change needed, so want to get some input on it before going full fledge on it.

Currently the PR:

 - Refactors `Owner` into `EthAddress` (dropping `Owner` and `OwnerInput` interfaces)
 - Introduces `HexString<Length>` and `NonPrefixedHexString<Length>` (see `./src/utils/hex.ts`)
 - Conversions utilities from/to hex string uses `HexString` (eq. prefixed) and is generally default internally (in opposing the `NonPrefixedHexString`)
 - Demos the approach of the refactor for some methods like `Bee.createFeedManifest()` and related functions

The most interesting parts of the PR:
 - `src/utils/hex.ts`
 - `src/feed/index.ts`
 - `src/bzz.ts`

Some limitations with the proposed implementation:
 - When you use utility functions like `bytesToHex` and you want to bound the length, you have to specify twice the length: `bytesToHex<6>(new Uint8Array([1,2]), 6)`. One adds compile-time information, the other checks it runtime. This might lead to the forgotten specifications of one of them.
 - Conversion to non-prefixed hex string using `stripHexPrefix<PrefixedL, NonPrefixedL>` requires to specify the lengths of the new non-prefixed hex string as TS can not do compile-time math (eq. subtract by two).

### Problem

From what I see, there is a bit of a mess between using `Bytes` and `HexString | string`. There is lot of `bytesToHex` and `hexToBytes` conversions around which I feel is creating quite some confusion (at least for me). **So I would like to propose basically minimizing the number of usages of these methods in the bee-js**

### Possible solutions

First of all, as we already discussed with @agazso and @vojtechsimetka on a call, there are two distinguished groups of data:

 1. content of the files/chunks etc. 
 2. references, hashes, signatures, addresses

The first group is for sure to stay in `Bytes` as it does not make sense to really represent it in other ways, I won't be considering this data for the rest of the PR. 
The second group though is a bit different as those are information with which people operate with rather often (passing/displaying addresses etc), a lot of libraries/tools pass around, and also the API calls often expect these data encoded.

I believe that the approach "convert everything to `Bytes`" might have been thanks to `Bytes<L>` supported the length bound back then and `HexString` did not. With this PR this won't be the case anymore, so I feel this decision could be revisited.

Few remarks:
 - API calls expect these data hex-encoded
 - most of the ecosystem is used to pass these data around in Hex strings 
 - only in higher-level functionality (e.g. feeds) we need actual bytes (e.g. for hashing)
 
#### 1. Adopt Hex strings as the internal data type

This solution would basically lead to drop the `Bytes<L>` type in most of its places and replace it with `HexString<L>` (length bound) by default (prefixed one). And only if necessary convert it to bytes locally.

This would require proper validation of the user's input together (that should happen anyway though) with type length specification.

#### 2. Keep `Bytes`, but...

This would keep the main internal data structure as `Bytes`, but... 
In my current view, this is not the case at the moment, or at least not fully. For example, the `./src/modules/*` files usually take `string` as a parameter (for Reference, topic, etc), which shows that currently the codebase actually uses both data types. 

If we would like to continue having `Bytes` internally, then we should push this usage to the end and use it everywhere, where only before making API requests actually convert those to hex if needed.


AuHau:
Closing in favor of #213

PR #203, opened 2021-03-11
fix: strip trailining slash in node url

### LINKED ISSUE #165: Check for trailing `/` in `url` in constructor
This might lead to weird errors for example in the browser where the browser might do [CORS Preflight check](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#preflighted_requests) (eq. `OPTIONS` request to Bee) to URL `http://localhost:1633//bytes` to which Bee replies with redirect response which browser rejects as an invalid reply.

AuHau:
Closes #165

### DIFF
--- a/test/bee-class.spec.ts
+++ b/test/bee-class.spec.ts
@@ -1,4 +1,4 @@
-import { Bee, BeeDebug, Collection } from '../src'
+import { Bee, BeeArgumentError, BeeDebug, Collection } from '../src'
 import { ChunkReference } from '../src/feed'
 import { HEX_REFERENCE_LENGTH } from '../src/types'
 import { makeBytes } from '../src/utils/bytes'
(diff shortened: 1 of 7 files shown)


test/bee-class.spec.ts
```diff
@@ -24,6 +24,29 @@ describe('Bee class', () => {
...
+    testUrl(undefined)
+    testUrl(1)
+    testUrl(Symbol.for('hello'))
+    testUrl('some-invalid-url')
+    testUrl('invalid:protocol')
+    // eslint-disable-next-line no-script-url
+    testUrl('javascript:console.log()')
+    testUrl('ws://localhost:1633')
+  })
+
```
agazso, line 49: It would be nice to have a test for actually stripping the trailing slash 😄

vojtechsimetka:
I'd like to point out one caveat which until now we tried to avoid. Creating instance of Bee or BeeDebug can now throw error. It is way better than failing on next use of the library, but is something we should properly documented (e.g. in tutorials).

test/bee-class.spec.ts
AuHau: Haha yeah good point :-D

AuHau:
Good point Vojtech. 

Sorry guys I did not finish it as I run into the problem with asserting the thrown errors and forgot to turn it into a draft. Will solve that and then make it ready for review again.

agazso (approved):

PR #207, opened 2021-03-15
feat: missing endpoints on Bee and BeeDebug class

vojtechsimetka:


src/modules/health.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Periodically ping the base bee URL
+ * emit 'check' event each time the ping happens
+ * emit 'error' with the Error as payload, if the endpoint is unreachable
+ *
+ * @param url        Bee URL
+ * @param frequency  How frequently should the health endpoint be pinged
+ */
+export function healthSubscribe(url: string, frequency = 1000): HealthEmitter {
+  const emitter = new Emittery.Typed<HealthEmitterData, HealthEmitterNonData>()
```
AuHau, line 27: Use the latest version of `Emittery` which [drops the the `.Typed<>` alias](https://github.com/sindresorhus/emittery/releases/tag/v0.8.0). We should stay with the latest version

src/bee.ts
```diff
@@ -462,4 +464,22 @@ export class Bee {
       upload: uploadSingleOwnerChunkData.bind(null, this.url, canonicalSigner),
     }
   }
+
+  /**
+   * Ping the base bee URL. If connection was not successful throw error
+   */
+  health(): Promise<void> | never {
```
AuHau, line 471: Hmmm, I find the usage of exceptions for this bit weird and I would say maybe counterintuitive. In my mind exceptions are something that happens only if something goes wrong unexpectedly. Here I very much expect that the node can be down. 

Also, I feel that this pretty much pushes the usage of `try/catch` on to the users as they might not want their app to crash because of this call and if they won't do proper checks they might catch actually some other valid exceptions (not sure if there is some to be expected though).

I would advise returning simple `boolean`.

src/bee.ts
```diff
@@ -462,4 +464,22 @@ export class Bee {
...
+    return health.health(this.url)
+  }
+
+  /**
+   * Periodically ping the base bee URL
+   * emit 'check' event each time the ping happens
```
AuHau, line 477:
```suggestion
   * @emits 'check' event each time the ping happens
```

There is `@emits/@fires` js doc directive ;-)

src/bee.ts
```diff
@@ -462,4 +464,22 @@ export class Bee {
...
+   * emit 'error' with the Error as payload, if the endpoint is unreachable
```
AuHau, line 478:
```suggestion
   * @emits 'error' with the Error as payload, if the endpoint is unreachable
```

AuHau (commented):
I am against throwing exception in the `health()`.

Regarding the new dependency, I am fine with it. It is a well-established library with a size 1,5 kB minified and in browser we would have to polyfill EventEmitter anyway.

Another option could be using async generators for this if we would like to avoid the extra dependency, but not sure how user friendly that would be.

src/bee.ts
agazso, line 471: On the Bee class the function names so far were imperative verbs (e.g. `downloadFile`, `createTag` etc.). For the sake of consistency consider using something like `checkConnection`. Alternatively if using a boolean for return value this could be called `isConnected`.

src/modules/health.ts
```diff
@@ -0,0 +1,39 @@
...
+    try {
+      await health(url)
+      emitter.emit('check')
+    } catch (e) {
+      emitter.emit('error', e)
+      clearTimeout(ref)
+    }
+  }, frequency)
+
+  return emitter
```
nugaon, line 38: I don't get this part, does it postpone the fetching of the EP then the process ends?
Was it intended to be called by `setInterval` instead of `setTimeout`?
If so, I think this function can be easily used in a wrong way by calling many times thereby generating many infinity loops during the session. I could imagine this in a singleton way instead of this one.

src/bee-debug.ts
```diff
@@ -149,7 +177,7 @@ export class BeeDebug {
    *
    * @param amount  Amount of tokens to deposit
    */
-  depositTokens(amount: number): Promise<DepositTokensResponse> {
+  depositTokens(amount: BigInt): Promise<DepositTokensResponse> {
```
agazso, line 180: Please be aware that `BigInt` is not supported on older browsers. This may be not an issue with an application (where you can display a page if the browser is not supported) but can be problematic in a library.

I usually use caniuse.com and as a rule of thumb I consider something well supported if it is supported more than 90% globally. In the case of `BigInt` this is only 88.72% currently, so it can be argued that it's almost there.

https://caniuse.com/?search=bigint

I don't think this is too important right now but maybe it's worth opening an issue about it once this is merged.

agazso (approved):
Looks good

Issue #143, opened 2021-02-23
Consider using FlavoredType instead of BrandedType for HexStrings


agazso:
See the discussion https://github.com/ethersphere/bee-js/pull/136#discussion_r579313832

PR #213, opened 2021-03-17
feat!: hex string length support

AuHau:
Supersedes #201 PR.

Breaking changes:
 - `HexString` type is considered as non-prefixed hex string and all functions using it are expecting to have it like that. If used on public API, then it should be sanitized with `makeHexString` function
 - dropped `stripHexPrefix` in favor of `makeHexString`
 - Hex utility functions no longer take `withPrefix: boolean` flag as they always output non-prefixed hex and instead take `len?: number` to assert expected length of the non-prefixed hex string
 - `Topic` type was changed from `Bytes` into `HexString`

### DIFF
--- a/src/modules/file.ts
+++ b/src/modules/file.ts
@@ -1,9 +1,10 @@
 import type { AxiosRequestConfig } from 'axios'
 import type { Readable } from 'stream'
-import { FileData, FileUploadOptions, UploadHeaders } from '../types'
+import { FileData, FileUploadOptions, Reference, UploadHeaders } from '../types'
 import { prepareData } from '../utils/data'
 import { extractUploadHeaders, readFileHeaders } from '../utils/headers'
 import { safeAxios } from '../utils/safeAxios'
+import { HexString } from '../utils/hex'
 
 const endpoint = '/files'
 
@@ -35,8 +36,8 @@ export async function upload(
   data: string | Uint8Array | Readable | ArrayBuffer,
   name?: string,
   options?: FileUploadOptions,
-): Promise<string> {
-  const response = await safeAxios<{ reference: string }>({
+): Promise<Reference> {
+  const response = await safeAxios<{ reference: Reference }>({
     ...options?.axiosOptions,
     method: 'post',
     url: url + endpoint,
--- a/src/modules/file.ts
+++ b/src/modules/file.ts
@@ -4,7 +4,6 @@ import { FileData, FileUploadOptions, Reference, UploadHeaders } from '../types'
 import { prepareData } from '../utils/data'
 import { extractUploadHeaders, readFileHeaders } from '../utils/headers'
 import { safeAxios } from '../utils/safeAxios'
-import { HexString } from '../utils/hex'
 
 const endpoint = '/files'
 
--- a/src/utils/eth.ts
+++ b/src/utils/eth.ts
@@ -1,6 +1,6 @@
 import { keccak256, sha3_256 } from 'js-sha3'
 import { BrandedString } from '../types'
-import { HexString, hexToBytes, intToHex, isHexString, makeHexString, assertHexString } from './hex'
+import { HexString, hexToBytes, intToHex, makeHexString, assertHexString } from './hex'
 import { Bytes, verifyBytes } from './bytes'
 
 export type OverlayAddress = BrandedString<'OverlayAddress'>
--- a/src/utils/eth.ts
+++ b/src/utils/eth.ts
@@ -18,7 +18,7 @@ export function makeEthAddress(address: EthAddress | Uint8Array | string): EthAd
   } else if (address instanceof Uint8Array) {
     return verifyBytes(ETH_ADDR_BYTES_LENGTH, address)
   }
-  throw new TypeError('invalid owner')
+  throw new TypeError('Invalid EthAddress')
 }
 
 export function makeHexEthAddress(address: EthAddress | Uint8Array | string): HexEthAddress {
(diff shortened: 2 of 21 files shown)


test/utils/hex.spec.ts
```diff
@@ -1,76 +1,72 @@
...
+      expect(isHexString('C0FFEE', 6)).toEqual(true)
+      expect(isHexString('C0FFEE', 7)).toEqual(false)
     })
-    test('empty string', () => {
-      const result = isHexString('')
 
-      expect(result).toBeFalsy()
-    })
-    test('chequebookaddress', () => {
-      const input = '0x20d7855b548C71b69dA434D46187C336BDcef00F'
```
agazso, line 45: The actual `chequebookadress` returned from the [endpoint](https://docs.ethswarm.org/debug-api/#tag/Chequebook/paths/~1chequebook~1address/get) is hex prefixed, unlike what the documentation says. This test was added because we have another test that validates the reply of that endpoint with `isHexString`. 

It seems that these tests should use `isPrefixedHexString` instead.

AuHau: I see! Thx! I am just looking into failing tests and I am a bit confused why it fails 😁

src/utils/eth.ts
```diff
@@ -1,24 +1,38 @@
...
+    const hexAddr = makeHexString(address, ETH_ADDR_HEX_LENGTH)
+    const ownerBytes = hexToBytes<typeof ETH_ADDR_BYTES_LENGTH>(hexAddr)
 
-    return verifyBytes(20, ownerBytes)
+    return verifyBytes(ETH_ADDR_BYTES_LENGTH, ownerBytes)
   } else if (address instanceof Uint8Array) {
-    return verifyBytes(20, address)
+    return verifyBytes(ETH_ADDR_BYTES_LENGTH, address)
   }
   throw new TypeError('invalid owner')
```
agazso, line 21: should be 'invalid eth address`

src/utils/eth.ts
```diff
@@ -1,24 +1,38 @@
...
 }
 
+export function makeHexEthAddress(address: EthAddress | Uint8Array | string): HexEthAddress {
+  try {
+    return makeHexString(address, ETH_ADDR_HEX_LENGTH)
+  } catch (e) {
+    if (e instanceof TypeError) {
+      e.message = `Invalid Owner: ${e.message}`
```
agazso, line 29: should be 'invalid eth address`

src/modules/file.ts
```diff
@@ -1,9 +1,10 @@
 import type { AxiosRequestConfig } from 'axios'
 import type { Readable } from 'stream'
-import { FileData, FileUploadOptions, UploadHeaders } from '../types'
+import { FileData, FileUploadOptions, Reference, UploadHeaders } from '../types'
 import { prepareData } from '../utils/data'
 import { extractUploadHeaders, readFileHeaders } from '../utils/headers'
 import { safeAxios } from '../utils/safeAxios'
+import { HexString } from '../utils/hex'
```
agazso, line 7: This can be removed

src/utils/eth.ts
```diff
@@ -37,23 +51,30 @@ function isEthAddrCaseIns(address: string | HexString | HexEthAddress): address
...
+
+    return true
+  } catch (e) {
+    if (e instanceof TypeError) {
       return false
     }
-  }
 
-  return true
+    throw e
```
agazso, line 76: I find this a bit strange that a type guard function may throw an exception. Can you describe what is the rationale here?

src/utils/hex.ts
```diff
@@ -1,90 +1,186 @@
...
+  if (typeof input === 'number') {
+    return intToHex<L>(input, len)
+  }
+
+  if (input instanceof Uint8Array) {
+    return bytesToHex<L>(input, len)
+  }
+
+  if (typeof input === 'string') {
+    try {
```
agazso, line 42: I find the control flow a bit convoluted with using the `try...catch` as a kind of conditional. It seems that it can be replaced with
```typescript
if (isPrefixedHexString(input)) {
  // ...
} else {
  assertHexString<L>(input, len)
  
  return input
}
```

src/utils/hex.ts
```diff
@@ -1,90 +1,186 @@
...
  */
-export function hexToBytes(hex: HexString): Uint8Array {
-  const hexWithoutPrefix = stripHexPrefix(hex)
-  const bytes = new Uint8Array(hexWithoutPrefix.length / 2)
+export function hexToBytes<Length extends number, LengthHex extends number = number>(
+  hex: HexString<LengthHex>,
+): Bytes<Length> {
+  assertHexString(hex)
+
+  const bytes = new Uint8Array(hex.length / 2)
```
agazso, line 77: Consider using `makeBytes(hex.length / 2)`

src/utils/eth.ts
AuHau: Well, generally speaking, you always want to catch only those errors that you know you can handle, which in this case is `TypeError` coming from `makeHexString` as it signals non valid hex string. Because if you will just "catch all" you can catch an error that is unrelated and the problem causing it might be silenced and then very hard to find.

If I would want to be more thorough I would actually check that it is coming from `makeHexString` (for example with error codes) as there might be more points on the way which could throw `TypeError`, but I found that overkill.

I agree that it might not be that much necessary in this case, but I find it generally a best practice to follow this style to prevent possible future problems.

agazso: In general this is a good practice, I agree.

However my assumption would be that a type guard is a pure function that returns true or false and it's written in a way that it can positively decide if the input is of a certain type. Every other cases would be false (including potential errors) because we already checked the positive cases.

agazso (approved):

nugaon (approved):
NIce work!

I made some comments which is related to ticket #143 where type problems has been stated and as I see this PR also brings in this flaw. We must work on that refactoring after this PR merge.

PR #215, opened 2021-03-18
chore: simplify tsconfigs

### LINKED ISSUE #209: Revisiting `tsconfig*` files for clarity
The `tsconfig*` could be condensed and simplified. Let's see if we can do something about that.

AuHau:
Closes #209 

This PR does two related things:

 1. fixes the webpack typings issues which lead to excluding `node_modules` and using `skipLibCheck` flag
 1. simplify the `tsconfig.json` setup with now having only one config which works for all our needed cases

I have tested the whole build pipeline and documentation generation and it works as expected.


AuHau:
FYI if you will try out the PR then don't forget to run `npm ci`! Otherwise you might get some compilation problems...

nugaon (approved):
LGTM!

@vojtechsimetka can you check those issues does not appear again that we had before with `typedoc`?

vojtechsimetka (approved):
> @vojtechsimetka can you check those issues does not appear again that we had before with `typedoc`?

I tried locally and it works fine. Also on the CI (see the Build docs part) https://github.com/ethersphere/bee-js/pull/215/checks?check_run_id=2137974792

Issue #202, opened 2021-03-11
Input validation of exposed funtions/methods


vojtechsimetka:
**Rationale**
As a developer and user of bee-js I would like to be alerted if I am using the functionality wrong = passing not supported or unexpected values. While we can not guarantee that the usage is correct, we can at least limit the potential wrongs by thorough type checking. This will also likely catch at least some security issues like passing functions that are evaluated in the code.

**Task**
Review all the exposed functions and methods and ensure all inputs are validated

**Concerns**
Argument: _Computation overhead_
Response: Type checking is generally not all that resource intensive (we are talking couple of instructions) if done right. Also, I assume most of the functions in `bee-js` are not going to be frequently invoked (e.g. at most every few seconds) so this validation code will not be run frequently.

nugaon:
In my opinion, the array checking - especially in case of big sized arrays - we shouldn't check _every_ element of the object in question.
E.g. https://github.com/ethersphere/bee-js/blob/a88277db374b3bafe273614954d43afad72018e8/src/utils/data.ts#L20
While you may be right about these functions won't be called frequently (though there could be a use-case always where this statement is false) these computational overheads don't yield better UX experience, because they seem like wasteful operations.
Nevertheless I support the idea the check every passed parameters in this reasonable sense.
Would be nice, if the PR which resolves this issue, also would fix #143 as well

AuHau:
Concrete problem as reported by @nugaon:

>I got 500 error on file upload if I pass File type (without array) for fileList parameter of uploadFile
It awaits array of files, but it does not check whether the user input is correct or not.

Moved here from: https://github.com/ethersphere/bee-js/issues/243

PR #217, opened 2021-03-19
refactor(chequebook): added integer validation and BigInt as type

vojtechsimetka:
`Number.MAX_SAFE_INTEGER` is `9007199254740991 ~= 9 * 10^16` which is just `9 BZZ`. It is reasonable to assume, that sooner or later users would want to withdraw or deposit more.

I've decided to extract this part from the #207 PR - see https://github.com/ethersphere/bee-js/pull/207#discussion_r597578613


agazso:
I just realized that we only use `BigInt` as an input type here. If we used `BigInt | number` then it would work the same as before (it wouldn't be a breaking change) but it could also work well with numbers bigger than max integer, no?

vojtechsimetka:
yeah but that could lead people to inocorrect usage. Take for an example a value just 10* bigger than `MAX_SAFE_INTEGER` and try to console log it:
```
console.log(90071992547409910)
// result   90071992547409900

console.log(90071992547409910n)
// result   90071992547409910n
```

agazso:
> yeah but that could lead people to inocorrect usage. Take for an example a value just 10* bigger than `MAX_SAFE_INTEGER` and try to console log it:
> 
> ```
> console.log(90071992547409910)
> // result   90071992547409900
> 
> console.log(90071992547409910n)
> // result   90071992547409910n
> ```

That's a good point and we can do [input validation](https://github.com/ethersphere/bee-js/issues/202) for the number type (also check for `Number.isInteger()`), which was missing previously btw.

vojtechsimetka:
@agazso are you still proposing to keep the `number` as input option but then have the following check?

```ts
isNumberInteger(amount: unknown) {
  return typeof amount === 'number'
      && amount > 0
      && amount < Number.MAX_SAFE_INTEGER
      && Number.isInteger(amount)
}

isBigInt(amount: unknown) {
  return typeof amount === 'bigint' && amount > 0)
}
```

agazso:
> @agazso are you still proposing to keep the `number` as input option but then have the following check? This would then be a runtime check of corse
> 
> ```ts
> isNumberInteger(amount: unknown) {
>   return typeof amount === 'number'
>       && amount > 0
>       && amount < Number.MAX_SAFE_INTEGER
>       && Number.isInteger(amount)
> }
> 
> isBigInt(amount: unknown) {
>   return typeof amount === 'bigint' && amount > 0
> }
> ```

Yes, something along these lines (input validation is a runtime check generally). Only difference that I would do the `amount > 0`  check separately, e.g. `(isNumberInteger(amount) || isBigInt(amount)) && amount > 0)`

src/utils/type.ts
```diff
@@ -0,0 +1,13 @@
...
+    typeof type === 'bigint' ||
+    (typeof type === 'number' &&
+      type > Number.MIN_SAFE_INTEGER &&
+      type < Number.MAX_SAFE_INTEGER &&
+      Number.isInteger(type))
+  )
+}
+
+export function assertInteger(type: unknown): asserts type is number | BigInt {
+  if (isInteger(type)) throw new TypeError('value is not integer')
```
agazso, line 12: This doesn't look correct 😄
It seems that a logical negation is missing.

src/utils/type.ts
```diff
@@ -0,0 +1,13 @@
+export function isInteger(type: unknown): type is number | BigInt {
```
agazso, line 1: I think `value` is a more appropriate name instead of `type`

agazso (changes requested):
`assertInteger` seems to do the opposite what its name suggests. Luckily the tests caught it 😄 

Other than that it looks good.

agazso (approved):

PR #219, opened 2021-03-19
feat: data helpers for downloaded bytes

### LINKED ISSUE #191: Return data type for download functions
Currently we are returning `Uint8Array` types from the download functions (and pssReceive) because it is suitable type for representing potentially binary data. However it is not easy to use and most people are not familiar with it. For example to convert it to string you have to use the `TextDecoder` class which is also something most developers are not familiar with (for illustration see #174).

Maybe it would be better to define a new type that could hold a `Uint8Array` for the binary data representation but would also provide conversion functions to text and json, similarly to the `Response` API used by `fetch`. For example:

```typescript
interface Data {
   readonly bytes: Uint8Array
   text: () => string
   json: () => Record<string, any>
}
```

Then we could return this interface from our functions everywhere with the help of an implementation that takes a `Uint8Array` and creates this type from it.

AuHau:
Closes #191 

Once approved, I will create follow up PR to `bee-js-docs` changing the docs.

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -64,6 +64,12 @@ export interface FileData<T> extends FileHeaders {
   data: T
 }
 
+export interface Data extends Uint8Array {
+  text: () => string
+  json: () => Record<string, any>
+  hex: () => HexString
+}
+
 /**
  * Object represents a file and some of its metadata in [[Directory]] object.
  */
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -68,11 +68,13 @@ export interface FileData<T> extends FileHeaders {
  * Helper interface that adds utility functions
  * to work more conveniently with bytes in normal
  * user scenarios.
+ *
+ * Concretely: text(), hex(), json()
  */
 export interface Data extends Uint8Array {
-  text: () => string
-  json: () => Record<string, any>
-  hex: () => HexString
+  text(): string
+  hex(): HexString
+  json(): Record<string, any>
 }
 
 /**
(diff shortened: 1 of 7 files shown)


src/types/index.ts
```diff
@@ -64,6 +64,12 @@ export interface FileData<T> extends FileHeaders {
   data: T
 }
 
+export interface Data extends Uint8Array {
+  text: () => string
+  json: () => Record<string, any>
```
agazso, line 69: Currently using `any` gives a warning and it seems that there is some space for improvements here regarding the typing. I could even imagine having an optional validator function passed in to provide a return type information and to ensure that the returned type is valid. It doesn't have to be in this PR, just something to keep in mind.

src/utils/bytes.ts
```diff
@@ -139,3 +142,11 @@ export function verifyBytesAtOffset<Length extends number>(
 ): Bytes<Length> {
   return verifyBytes(length, bytesAtOffset(offset, length, data))
 }
+
+export function wrapBytesWithHelpers(data: Uint8Array): Data {
+  return Object.assign(data, {
+    text: () => new TextDecoder('utf-8').decode(data),
+    json: () => JSON.parse(new TextDecoder().decode(data)),
```
agazso, line 149: The default encoding for `TextDecoder` is `utf-8`, so specifying it is not necessary. If you want to keep it then I suggest to add it to here as well.

https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder

agazso (approved):
👍

PR #223, opened 2021-03-24
feat: improved signer interface and validation

AuHau:
So I did not take the road of breaking change only expanded the flexibility of the interface and improved validation of custom Signers.

I will create follow up PR building on this one with utility for wrapping the Ethereum wallets.

### DIFF
--- a/src/chunk/signer.ts
+++ b/src/chunk/signer.ts
@@ -168,7 +168,7 @@ export async function sign(signer: Signer, data: Uint8Array): Promise<Signature>
   const result = await signer.sign(wrapBytesWithHelpers(data))
 
   if (typeof result === 'string') {
-    assertHexString(result, SIGNATURE_HEX_LENGTH)
+    makeHexString(result, SIGNATURE_HEX_LENGTH)
 
     return hexToBytes<65>(result)
   }
--- a/src/chunk/signer.ts
+++ b/src/chunk/signer.ts
@@ -2,7 +2,7 @@ import { ec, curve } from 'elliptic'
 import { BeeError } from '../utils/error'
 import { Bytes, isBytes, verifyBytes, wrapBytesWithHelpers } from '../utils/bytes'
 import { keccak256Hash } from './hash'
-import { assertHexString, HexString, hexToBytes, makeHexString } from '../utils/hex'
+import { HexString, hexToBytes, makeHexString } from '../utils/hex'
 import { EthAddress } from '../utils/eth'
 import { Data } from '../types'
 
@@ -41,7 +41,7 @@ type EllipticPublicKey = curve.base.BasePoint
  * For example see the hashWithEthereumPrefix() function.
  *
  * @property sign     The sign function that can be sync or async. This function takes non-prefixed data. See above.
- * @property address  The ethereum address of the signer in bytes or hex string.
+ * @property address  The ethereum address of the signer in bytes.
  * @see hashWithEthereumPrefix
  */
 export type Signer = {
--- a/src/chunk/signer.ts
+++ b/src/chunk/signer.ts
@@ -23,8 +23,8 @@ export type PublicKey = Bytes<32> | Bytes<64>
  *
  * @see Data
  */
-type SyncSigner = (digest: Data) => Signature | HexString<typeof SIGNATURE_HEX_LENGTH>
-type AsyncSigner = (digest: Data) => Promise<Signature | HexString<typeof SIGNATURE_HEX_LENGTH>>
+type SyncSigner = (digest: Data) => Signature | HexString<typeof SIGNATURE_HEX_LENGTH> | string
+type AsyncSigner = (digest: Data) => Promise<Signature | HexString<typeof SIGNATURE_HEX_LENGTH> | string>
 type EllipticPublicKey = curve.base.BasePoint
 
 /**
--- a/src/chunk/signer.ts
+++ b/src/chunk/signer.ts
@@ -1,20 +1,31 @@
 import { ec, curve } from 'elliptic'
 import { BeeError } from '../utils/error'
-import { Bytes, verifyBytes } from '../utils/bytes'
+import { Bytes, isBytes, verifyBytes, wrapBytesWithHelpers } from '../utils/bytes'
 import { keccak256Hash } from './hash'
-import { hexToBytes, makeHexString } from '../utils/hex'
+import { assertHexString, HexString, hexToBytes, makeHexString } from '../utils/hex'
 import { EthAddress } from '../utils/eth'
+import { Data } from '../types'
 
 /**
  * Ethereum compatible signing and recovery
  */
+const SIGNATURE_HEX_LENGTH = 130
+const SIGNATURE_BYTES_LENGTH = 65
 
-export type Signature = Bytes<65>
+export type Signature = Bytes<typeof SIGNATURE_BYTES_LENGTH>
 export type PrivateKey = Bytes<32>
 export type PublicKey = Bytes<32> | Bytes<64>
 
-type SyncSigner = (digest: Uint8Array) => Signature
-type AsyncSigner = (digest: Uint8Array) => Promise<Signature>
+/**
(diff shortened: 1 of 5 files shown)


agazso:
There is a broken test: https://github.com/ethersphere/bee-js/pull/223/checks?check_run_id=2182134730#step:11:365

AuHau:
Yeah working on it. I was finishing #224

src/chunk/signer.ts
```diff
@@ -136,8 +157,27 @@ export function makeSigner(signer: Signer | Uint8Array | string | unknown): Sign
...
+  assertSigner(signer)
+
+  return signer
+}
+
+export async function sign(signer: Signer, data: Uint8Array): Promise<Signature> {
+  const result = await signer.sign(wrapBytesWithHelpers(data))
+
+  if (typeof result === 'string') {
+    assertHexString(result, SIGNATURE_HEX_LENGTH)
```
agazso, line 171: By default the returned signature from ethereum is prefixed hex so I think it would be good to add support for that use-case. Otherwise the user has to strip the prefix. 

Also the type we expect here is`HexString` but for someone who is not familiar with our internal naming conventions (which can be assumed by default) this can be confusing.

AuHau: Very good point, thanks for spotting that!

src/chunk/signer.ts
```diff
@@ -136,8 +157,27 @@ export function makeSigner(signer: Signer | Uint8Array | string | unknown): Sign
...
+  assertSigner(signer)
+
+  return signer
+}
+
+export async function sign(signer: Signer, data: Uint8Array): Promise<Signature> {
+  const result = await signer.sign(wrapBytesWithHelpers(data))
+
+  if (typeof result === 'string') {
+    makeHexString(result, SIGNATURE_HEX_LENGTH)
```
agazso, line 171: `makeHexString` returns a new value, but the return value here is discarded, so this won't work correctly for unprefixed hex strings.

It would be good to write tests for both the prefixed and unprefixed cases, and also an invalid hex string case.

AuHau: Ups 🙈 Thanks! 😅

src/chunk/signer.ts
```diff
@@ -25,12 +36,12 @@ type AsyncSigner = (digest: Uint8Array) => Promise<Signature>
...
  * to do this prefixing manually if you use the `personal_sign` method. Check documentation of the tool!
  * If you are writing your own storage for keys, then you have to prefix the data manually otherwise the Bee node
- * will reject the chunk!
+ * will reject the chunks signed by you!
  *
  * For example see the hashWithEthereumPrefix() function.
  *
  * @property sign     The sign function that can be sync or async. This function takes non-prefixed data. See above.
- * @property address  The ethereum address of the signer
+ * @property address  The ethereum address of the signer in bytes or hex string.
```
agazso, line 44: The hex address support has not been implemented, right?

src/chunk/signer.ts
```diff
@@ -1,20 +1,31 @@
 import { ec, curve } from 'elliptic'
 import { BeeError } from '../utils/error'
-import { Bytes, verifyBytes } from '../utils/bytes'
+import { Bytes, isBytes, verifyBytes, wrapBytesWithHelpers } from '../utils/bytes'
 import { keccak256Hash } from './hash'
-import { hexToBytes, makeHexString } from '../utils/hex'
+import { assertHexString, HexString, hexToBytes, makeHexString } from '../utils/hex'
```
agazso, line 5: Please remove unused `assertHexString`

agazso (approved):
Looks good, see my other comments

src/chunk/signer.ts
AuHau: Yeah left over from my playing with this approach. Thanks!

src/chunk/signer.ts
AuHau: Yeah will do. Thanks

src/chunk/signer.ts
```diff
@@ -117,13 +126,25 @@
...
+  if (!isBytes(20, typedSigner.address)) {
+    throw new TypeError("Signer's address must be Uint8Array with 20 bytes!")
+  }
+
+  if (typeof typedSigner.sign !== 'function') {
+    throw new TypeError('Signer sign property needs to be function!')
+  }
 }
 
 export function makeSigner(signer: Signer | Uint8Array | string | unknown): Signer {
```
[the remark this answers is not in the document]
nugaon: Also you can remove `unknown` from here

test/chunk/signer.spec.ts
```diff
@@ -63,4 +78,59 @@ describe('signer', () => {
...
+          return expectedSignatureHex
+        },
+        address: makeBytes(20),
+      } as Signer
+
+      const result = await sign(signer, dataToSignBytes)
+      expect(result).toEqual(expectedSignatureBytes)
+    })
+
+    function testSignerConversion(input: string, output: Uint8Array): void {
```
nugaon, line 99: I would change `string` to the corresponding type that it awaits in order to be more readable the test and don't do the assertion later on the `signer` function

test/chunk/signer.spec.ts
```diff
@@ -63,4 +78,59 @@ describe('signer', () => {
...
+      it(`should convert sign result ${shorten(input)}`, async () => {
+        const signer = {
+          sign: digest => {
```
nugaon, line 102:
```suggestion
          sign: () => {
```

test/chunk/signer.spec.ts
```diff
@@ -63,4 +78,59 @@ describe('signer', () => {
...
+      })
+    }
+
+    testSignerConversion(expectedSignatureHex, expectedSignatureBytes)
+    testSignerConversion(`0x${expectedSignatureHex}`, expectedSignatureBytes)
+
+    function testSignerThrowing(input: unknown): void {
+      it(`should throw for invalid result ${shorten(input)}`, async () => {
+        const signer = {
+          sign: digest => {
```
nugaon, line 119:
```suggestion
          sign: () => {
```

src/chunk/signer.ts
```diff
@@ -136,8 +157,27 @@
     const verifiedPrivateKey = verifyBytes(32, signer)
 
     return makeDefaultSigner(verifiedPrivateKey)
```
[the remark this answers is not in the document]
AuHau: It is already done in the other PR.

src/chunk/signer.ts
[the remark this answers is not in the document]
AuHau: Well, this is a structure from Attila, but to my understanding `make*` should be a "factory" that from multiple compatible inputs should construct that desired data structure or throw an error. From this definition, it should be clear that such a structure should not error out when you pass it already they desired data structure so yes there is also `Signer` interface part of the input data types.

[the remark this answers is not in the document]
agazso: This was called `verifySigner` before and as such it made more sense to accept the `Signer` type as well. I am not sure what could be a good convention here, but please keep in mind that this function is used mainly for input verification, so that it accepts all possible input types from the user and returns a signer object or throws error.

It would be possible to split this into separate `isSigner` check and then a `makeSigner` call but then it would make the input validation more complex, because you would have to copy that logic everywhere where we accept signers from the user.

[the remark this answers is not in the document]
nugaon: I don't see users who have already a signer why would pass that to this function. I think we could split this functionality as not really intuitive.
I checked and there are two references to this `makeSigner` originated from `Bee` class.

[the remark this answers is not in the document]
agazso: Soon that is going to be three: https://github.com/ethersphere/bee-js/pull/224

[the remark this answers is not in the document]
AuHau: Hmm while I think I understand your point @nugaon, I don't really think it is confusing that much and hence I am more with @agazso on this.

PR #224, opened 2021-03-24
feat: default signer on bee instance

AuHau:
Introduce default signer for Bee instance.

### DIFF
--- a/src/bee.ts
+++ b/src/bee.ts
@@ -19,6 +19,7 @@ import type {
   CollectionUploadOptions,
   FileUploadOptions,
   Data,
+  BeeOptions,
 } from './types'
 import { BeeError } from './utils/error'
 import { prepareWebsocketData } from './utils/data'
@@ -38,26 +39,34 @@ import { assertReference } from './utils/type'
 
 /**
  * The Bee class provides a way of interacting with the Bee APIs based on the provided url
- *
- * @param url URL of a running Bee node
  */
 export class Bee {
   public readonly url: string
+  public readonly signer?: Signer
 
-  constructor(url: string) {
+  /**
+   * @param url URL of a running Bee node
+   * @param options
+   * @param options.signer Optional Signer object/string/Uint8Array that will be default signer for the instance.
+   */
+  constructor(url: string, options?: BeeOptions) {
     assertBeeUrl(url)
 
     // Remove last slash if present, as our endpoint strings starts with `/...`
     // which could lead to double slash in URL to which Bee responds with
     // unnecessary redirects.
     this.url = stripLastSlash(url)
+
+    if (options?.signer) {
+      this.signer = makeSigner(options.signer)
+    }
   }
 
   /**
    * Upload data to a Bee node
    *
    * @param data    Data to be uploaded
-   * @param options Aditional options like tag, encryption, pinning, content-type
+   * @param options Additional options like tag, encryption, pinning, content-type
    *
    * @returns reference is a content hash of the data
    */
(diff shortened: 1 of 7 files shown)


test/bee-class.spec.ts
```diff
@@ -350,4 +350,38 @@ describe('Bee class', () => {
...
+  describe('signer', () => {
+    const socHash = '9d453ebb73b2fedaaf44ceddcf7a0aa37f3e3d6453fea5841c31f0ea6d61dc85' as HexString
+
+    it('should be possible to pass it in constructor', async () => {
+      const identifier = makeBytes(32) // all zeroes
+      const socAddress = makeSOCAddress(identifier, makeEthAddress(testIdentity.address))
+      await tryDeleteChunkFromLocalStorage(socAddress)
+
+      const bee = new Bee(BEE_URL, { signer: testIdentity.privateKey })
+      const socWriter = bee.makeSOCWriter()
```
agazso, line 363: It would be good to have similar tests for `makeFeedWriter`

agazso (approved):
Looks good, I would add tests for the feed writers as well.

test/bee-class.spec.ts
AuHau: Hmm not sure what value would that bring. I am sort of "abusing" the `socWriter` here to verify that the correct signer is used with it. The main tested logic is directly in the `Bee` class and is shared between both SOC and Feed logic (the `resolveSigner()` method).

agazso: Ideally with the changes in this PR these are the potential cases to be tested when calling `makeSOCWriter`:
- no signer in constructor, no signer provided as argument -> error
- signer in constructor, no signer provided as argument -> signer in constructor called
- singer in constructor, signer provided as argument -> signer from argument called

And similarly these cases tested with `makeFeedWriter`.

AuHau: Well, exactly these test cases are already present. I don't think it make much sense to test it also with `makeFeedWriter` but it is not problem for me, so I will add these.

src/types/index.ts
```diff
@@ -1,6 +1,7 @@
 import { BeeError } from '../utils/error'
 import type { AxiosRequestConfig } from 'axios'
 import { HexString } from '../utils/hex'
+import { Signer } from '../chunk/signer'
```
nugaon, line 4:
```suggestion
import type { Signer } from '../chunk/signer'
```

src/bee.ts
```diff
@@ -38,26 +39,34 @@ import { assertReference } from './utils/type'
...
- *
- * @param url URL of a running Bee node
  */
 export class Bee {
   public readonly url: string
+  public readonly signer?: Signer
 
-  constructor(url: string) {
+  /**
+   * @param url URL of a running Bee node
```
nugaon, line 48: This is the Bee client's API URL, please correct this jsdoc.
In an other PR we need to make clear distinction between `apiUrl` and `debugUrl` in codes which face to devs.

src/bee.ts
```diff
@@ -38,26 +39,34 @@ import { assertReference } from './utils/type'
...
+   * @param options
+   * @param options.signer Optional Signer object/string/Uint8Array that will be default signer for the instance.
```
nugaon, line 50:
```suggestion
   * @param [options.signer] Signer object or private key of the Signer in form of either hex string or Uint8Array that will be default signer for the instance.
```

Please describe what exactly the object/string/Uint8Array in this context.

src/bee.ts
```diff
@@ -38,26 +39,34 @@ import { assertReference } from './utils/type'
...
- * @param url URL of a running Bee node
  */
 export class Bee {
   public readonly url: string
+  public readonly signer?: Signer
 
-  constructor(url: string) {
+  /**
+   * @param url URL of a running Bee node
+   * @param options
```
nugaon, line 49:
```suggestion
   * @param [options]
```

nugaon (approved):
I left some comments, please consider those, but seems good to me.

src/bee.ts
AuHau: What is the functionality of the square brackets?

Issue #38, opened 2020-12-11
Abstract tests in /modules to run with Bee class as well


vojtechsimetka:
**Tension:**
Right now we are unnecessarily replicating tests in the Bee class that are already in the individual modules

**Proposed solution:**
Abstract the tests so that they can be run on both the Bee class and the module. Good example how this can be done is https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/token/ERC777/ERC777.test.js#L43-L59 

See discussion https://github.com/ethersphere/bee-js/pull/15#discussion_r534837737

vojtechsimetka:
On the testing meeting on March 7th 2021 we decided we do not need to run the tests against both Bee class and individual modules. We will only run the tests for the functionality against the modules themselves (just as we do right now) and then test the bee class only for the assertion of params #248 .

PR #228, opened 2021-03-25
feat!: replaced `getPssPublicKey` with `getNodeAddresses`

vojtechsimetka:


agazso (commented):
I know that there were no prior tests (and it's my fault 😅 ), but maybe it would be a good idea to add tests, at least when we introduce breaking changes.

vojtechsimetka:
@agazso  This brings us again to the discussion of having tests that run against both the `modules` and the `Bee`/`BeeDebug` class itself #38 . Because we do test this function in the module :) https://github.com/ethersphere/bee-js/blob/master/test/modules/debug/connectivity.spec.ts#L54-L62

AuHau (approved):
LGTM

nugaon (approved):

PR #229, opened 2021-03-25
ci: using specific version of the bee node for tests

vojtechsimetka:
Seems that in past we always used `latest` tag as the `sed` did not work. I introduced this with #193


AuHau (approved):
Hmm, interesting! LGTM.

nugaon (approved):

nugaon:
Maybe would be good to apply that practice to print the version number of all used tools/apps within a test workflow.

PR #168, opened 2021-03-02
docs: eth wallet signing example

AuHau:
Waits for release of #194

PR #230, opened 2021-03-26
feat: ethereum wallet signer utility

AuHau:
Currently, the target of the PR is `feat/improved-signer` as I was not able to merge the original PR. I will rebase it once it is merged.

It is a draft as I will verify its functionality on the Metamask signing example first.

### DIFF
--- a/src/utils/eth.ts
+++ b/src/utils/eth.ts
@@ -218,7 +218,7 @@ export async function makeEthereumWalletSigner(
       const result = await executorFnc({
         jsonrpc: '2.0',
         method: 'personal_sign',
-        params: ['0x' + hexEthAddress, '0x' + data.hex()],
+        params: ['0x' + data.hex(), '0x' + hexEthAddress],
       })
 
       return result as string
--- a/src/utils/eth.ts
+++ b/src/utils/eth.ts
@@ -210,7 +210,7 @@ export async function createEthereumWalletSigner(
   }
 
   const bytesEthAddress = makeEthAddress(ethAddress)
-  ethAddress = makeHexEthAddress(ethAddress)
+  const hexEthAddress = makeHexEthAddress(ethAddress)
 
   return {
     address: bytesEthAddress,
@@ -218,7 +218,7 @@ export async function createEthereumWalletSigner(
       const result = await executorFnc({
         jsonrpc: '2.0',
         method: 'personal_sign',
-        params: ['0x' + ethAddress, '0x' + data.hex()],
+        params: ['0x' + hexEthAddress, '0x' + data.hex()],
       })
 
       return result as string
--- a/src/utils/eth.ts
+++ b/src/utils/eth.ts
@@ -1,7 +1,8 @@
 import { keccak256, sha3_256 } from 'js-sha3'
-import { BrandedString } from '../types'
+import { BrandedString, Data } from '../types'
 import { HexString, hexToBytes, intToHex, makeHexString, assertHexString } from './hex'
 import { Bytes, verifyBytes } from './bytes'
+import { Signer } from '../chunk/signer'
 
 export type OverlayAddress = BrandedString<'OverlayAddress'>
 export type EthAddress = Bytes<20>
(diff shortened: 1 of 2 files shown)


AuHau:
So the code was verified on the example! It is ready for review! :-)

AuHau:
If you want to see example how it is used see here: https://github.com/ethersphere/bee-js/pull/168/files#diff-2a4c36fe33bf013c275d694caafa6694f0736f0eb836c97b80dae3e29258ad3dR38

src/utils/eth.ts
```diff
@@ -167,3 +168,60 @@ export function ethToSwarmAddress(ethAddress: string | HexString | HexEthAddress
...
+  } else {
+    throw new Error('Incompatible interface of given provider!')
+  }
+
+  if (!ethAddress) {
+    ethAddress = ((await executorFnc({ method: 'eth_requestAccounts' })) as string[])[0]
+  }
+
+  const bytesEthAddress = makeEthAddress(ethAddress)
+  ethAddress = makeHexEthAddress(ethAddress)
```
agazso, line 213: Maybe it would be better to introduce a new variable (e.g. `hexEthAddress`) than overwriting an existing one for clarity and future maintainability.

src/utils/eth.ts
```diff
@@ -167,3 +168,60 @@ export function ethToSwarmAddress(ethAddress: string | HexString | HexEthAddress
...
+interface RequestArguments {
+  jsonrpc?: string
+  method: string
+  params?: unknown[] | Record<string, unknown>
+}
+
+export interface JsonRPC {
+  request?(args: RequestArguments): Promise<unknown>
+  sendAsync?(args: RequestArguments): Promise<unknown>
+}
```
vojtechsimetka, line 181: Shouldn't this be in `types/index.ts`?

vojtechsimetka (approved):
LGTM

src/utils/eth.ts
AuHau: I don't know. To me there is no clear guidelines on what to put to `types/index.ts` and what not. I guess the only thing I can think of is that public interfaces should be placed there, which I guess this is somewhat. I will move it there then.

src/utils/eth.ts
```diff
@@ -167,3 +168,60 @@ export function ethToSwarmAddress(ethAddress: string | HexString | HexEthAddress
...
+  const bytesEthAddress = makeEthAddress(ethAddress)
+  const hexEthAddress = makeHexEthAddress(ethAddress)
+
+  return {
+    address: bytesEthAddress,
+    sign: async (data: Data): Promise<string> => {
+      const result = await executorFnc({
+        jsonrpc: '2.0',
+        method: 'personal_sign',
+        params: ['0x' + hexEthAddress, '0x' + data.hex()],
```
nugaon, line 221: Everywhere I checked the `data` is the first and the eth address is the second parameter. in here why is it differently?
https://web3js.readthedocs.io/en/v1.2.11/web3-eth-personal.html#sign

src/utils/eth.ts
vojtechsimetka: yeah the rule of a thumb for me is - can user see any variable which is of this interface? If so then it goes into `types/index.ts` which exposes it.

nugaon:
I'm little bit sceptical, because the integration test is missing. Is there any idea how can we try out Metamask operations with our test environment? maybe we should integrate their extension to our puppeteer browser?

src/utils/eth.ts
agazso: As far as I understand this is provided in either by `web3` or the wallet library that you are using. We only defined this for ourselves so that we don't have any in our codebase. If that's the case then I wouldn't provide it to the users, because they would only use this when using a wallet library and they should rely on the definitions of that.

AuHau: Yeah it is as Attila mentioned and I am with him on this 👍

AuHau:
> I'm little bit sceptical, because the integration test is missing. Is there any idea how can we try out Metamask operations with our test environment? maybe we should integrate their extension to our puppeteer browser?

I don't know if there is a way how to do integration tests for it, but I am 99% sure it is not worthy of the time and I am against that. There gonna be example using it in the examples repo which can be used for verification of the functionality.

agazso:
> I don't know if there is a way how to do integration tests for it, but I am 99% sure it is not worthy of the time and I am against that. There gonna be example using it in the examples repo which can be used for verification of the functionality.

Is there a way to test this functionality currently?

AuHau:
Except of manual testing with the example as I have mentioned earlier, and stubbing the provider that I have done already? I don't know about a way. There might be some way how to load extension to pupeeter but I have never done it and I think it would be too time consuming.

agazso:
> Except of manual testing with the example as I have mentioned earlier, and stubbing the provider that I have done already? I don't know about a way. There might be some way how to load extension to pupeeter but I have never done it and I think it would be too time consuming.

I think manual testing is fine, but to me it's not clear how I could test it. Could you provide the instructions for that?

src/utils/eth.ts
AuHau: Yeah, I noticed this inconsistency as well, but this came from some part of Metamask documentation although I can't find where now. I have dig bit deeply and it seems that the "official" way is the one you link as per discussion in here: https://github.com/MetaMask/metamask-extension/issues/10297

So I will swap it. Metamask should handle both versions.

AuHau:
Checkout `examples/metamask-signing` branch and go `/examples/eth-wallet-signing/`, run `npm install && npm start` and there you have it :-)

nugaon:
> There might be some way how to load extension to pupeeter but I have never done it and I think it would be too time consuming.

Well from my prespective it is not very time consuming as I integrated unpacked extension in [swarm-extension](https://github.com/ethersphere/swarm-extension) repo.

But if you say you've tested everything manually then it's okay from my side as well now

src/utils/eth.ts
nugaon: from the issue it doesn't seem like it handles both versions  : D but yeah we have to swap it as I mentioned.

PR #233, opened 2021-03-27
test: increased timeouts for tests testing failure and big files

vojtechsimetka:
- increase the amount of nodes spawned by `test/bee.sh` script to 5 which matches the CI
- increased timeout for tests with big files (it just takes longer now for some reason - maybe worse performing CI machine)
- increased timeout for tests which test failure (e.g. downloading non existing file, pining non existing collection). All these seem to now hit some internal timeout and probably should be consulted with Bee devs

Mind you, the test are probably not run correctly against specific release. #229 should fix it but I wanted to keep these PRs separate

### DIFF
--- a/test/bee.sh
+++ b/test/bee.sh
@@ -56,7 +56,7 @@ log_queen() {
 
 # Init variables
 EPHEMERAL=false
-WORKERS=2
+WORKERS=4
 QUEEN_CONTAINER_NAME="bee-queen-test"
 WORKER_CONTAINER_NAME="bee-worker-test"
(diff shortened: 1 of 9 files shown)


AuHau (approved):
LGTM

test/bee.sh
```diff
@@ -56,7 +56,7 @@ log_queen() {
 
 # Init variables
 EPHEMERAL=false
-WORKERS=2
```
nugaon, line 59: please amend the help description as well under the `workers` option

nugaon (approved):

PR #234, opened 2021-03-29
refactor!: isEthAddress to isHexEthAddress

AuHau:


vojtechsimetka (approved):
LGTM. This is breaking change though as we expose the function, right? So please add `!` and merge as:
`refactor!: isEthAddress to isHexEthAddress`

AuHau:
Good point thanks!

PR #235, opened 2021-03-29
feat: expose external types

### LINKED ISSUE #214: Export all needed interfaces
Currently, we expose types from `src/types/index.ts`, but that file does not contain all types in the project. Some of the interfaces are spread around the modules. While a lot of them are internal, there are some that should be actually external. For example the `Signer` interface.

We should go over the codebase and see what interfaces are missing and either move them to `src/types/index.ts` or expose them in another way. This is also important from the documentation perspective as only exported elements are shown in the generated API docs.

AuHau:
Closes #214 

This moves several types spread around files into `types/index.ts` as this is exposed on root level and hence is accessible to user, but more importantly it is exposed to `typedoc` generation process so these types will be documented in API Reference.

### DIFF
--- a/src/bee.ts
+++ b/src/bee.ts
@@ -19,18 +19,23 @@ import type {
   CollectionUploadOptions,
   FileUploadOptions,
   Data,
+  Signer,
+  FeedReader,
+  FeedWriter,
+  SOCWriter,
+  SOCReader,
+  Topic,
   BeeOptions,
 } from './types'
 import { BeeError } from './utils/error'
 import { prepareWebsocketData } from './utils/data'
 import { fileArrayBuffer, isFile } from './utils/file'
 import { AxiosRequestConfig } from 'axios'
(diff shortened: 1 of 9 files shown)


vojtechsimetka (approved):

AuHau:
Hmm yeah, usually do that, but I used GitHub's "Resolve conflicts" feature which apparently does Merge... But I am wondering does it actually matter if we Squash the commit in the end?

nugaon:
> Hmm yeah, usually do that, but I used GitHub's "Resolve conflicts" feature which apparently does Merge... But I am wondering does it actually matter if we Squash the commit in the end?

it matters when I would like to review only your changes now

vojtechsimetka:
> it matters when I would like to review only your changes now

I agree rebase is better, but github does diff on the files so rebase or merge should yield the same result.

nugaon:
> I agree rebase is better, but github does diff on the files so rebase or merge should yield the same result.

I'm checking your PRs in VS

PR #236, opened 2021-03-29
feat: get all published Bee versions from github

vojtechsimetka:
We talked about this briefly today in daily so I decided to give it a try and implement it. Added two static functions to the BeeDebug 
 class:
```ts
BeeDebug.getBeeVersions() // returns array of all Bee versions published to github
BeeDebug.getBeeVersionLatest() // returns latest Bee version published to github
```

Let's discuss here, I'm sure this will be opinionated feature :)

### DIFF
--- a/src/bee-debug.ts
+++ b/src/bee-debug.ts
@@ -23,6 +23,7 @@ import type {
   PingResponse,
   Health,
   NodeAddresses,
+  BeeVersion,
 } from './types'
 import { assertBeeUrl, stripLastSlash } from './utils/url'
 import { assertInteger } from './utils/type'
@@ -42,6 +43,24 @@ export class BeeDebug {
     // which could lead to double slash in URL to which Bee responds with
     // unnecessary redirects.
     this.url = stripLastSlash(url)
+  }
+
+  /**
+   * Get all published Bee versions from github
+   *
+   * @returns List of published and released Bee versions
+   */
+  static getBeeVersions(): Promise<BeeVersion[]> {
+    return status.getBeeVersions()
+  }
+
+  /**
+   * Get latest released Bee versions from github
+   *
+   * @returns Latest released Bee version
+   */
+  static getBeeVersionLatest(): Promise<BeeVersion> {
+    return status.getBeeVersionLatest()
   }
 
   getNodeAddresses(): Promise<NodeAddresses> {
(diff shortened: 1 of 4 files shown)


src/bee-debug.ts
```diff
@@ -44,6 +45,24 @@ export class BeeDebug {
...
+  static getBeeVersions(): Promise<BeeVersion[]> {
+    return status.getBeeVersions()
+  }
+
+  /**
+   * Get latest released Bee versions from github
+   *
+   * @returns Latest released Bee version
+   */
+  static getBeeVersionLatest(): Promise<BeeVersion> {
```
agazso, line 62: I would prefer calling this `getLatestBeeVersion`

src/modules/debug/status.ts
```diff
@@ -1,10 +1,14 @@
 import { safeAxios } from '../../utils/safeAxios'
-import type { Health } from '../../types/debug'
+import type { Health, BeeVersion } from '../../types/debug'
+
+const GITHUB_API_URL = 'https://api.github.com/repos/ethersphere/bee-js'
```
agazso, line 4: It seems to me that this returns the `bee-js` versions, not the `bee` ones. If that is the intent maybe it would be better to call this functionality `getBeeJsVersion` instead of `getBeeVersion` because to me the latter suggests that it is the version of the connected Bee node.

agazso:
Can you please describe the use-case here? I don't recall from our meeting why this is needed to be in the `bee-js` library, but it would also be good to document.

src/modules/debug/status.ts
vojtechsimetka: Yes that is an error, thank you!

vojtechsimetka:
So the motivation comes really from the `bee-status` project where we want to check that the running Bee node version is the latest version and if not instruct the user to update it. Right now, this is the only call there that is made directly with axios and not through `bee-js` library.

Although I do agree this is not going to be the most utilised functionality, I do believe there will be a number of use cases where this should be used. E.g. monitoring and dashboard tools or package managers. It also makes even more sense with PR I am preparing next which would list the Bee versions that are compatible with given `bee-js` version.

AuHau:
I agree that it won't be the coolest feature we support, but I guess it can be handy. LGTM, but will leave the approval to others.

vojtechsimetka:
Closing as we agreed this should not be right now in the scope of `bee-js`

Issue #211, opened 2021-03-16
Use `assert...` instead of `verify` functions


AuHau:
From conversation around #71 PR there emerges guideline to use `assertSomething()` as a convention to verify data type/value also using TS's [`asserts`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) directive and type guard.

We should incrementaly drop out of the code current `verify...` functions in favor of these. For example `verifyBytes`.

PR #238, opened 2021-03-30
refactor: split soc and cac tests and drop verify*

### LINKED ISSUE #150: Split SOC and CAC tests
Currently, all the tests chunks (both SOC and CAC) are bundled in one file `soc.spec.ts` which is `describe()` to `soc`, but included also tests for `cac`. This should be split and named correctly.

AuHau:
Closes #150 also partly covers #211

### DIFF
--- a/src/chunk/cac.ts
+++ b/src/chunk/cac.ts
@@ -1,7 +1,7 @@
 import { BrandedType } from '../types'
 import { BeeError } from '../utils/error'
 import { bmtHash } from './bmt'
-import { Bytes, bytesAtOffset, bytesEqual, FlexBytes, flexBytesAtOffset, verifyFlexBytes } from '../utils/bytes'
+import { Bytes, bytesEqual, FlexBytes, flexBytesAtOffset, verifyFlexBytes } from '../utils/bytes'
 import { serializeBytes } from './serialize'
 import { makeSpan, SPAN_SIZE } from './span'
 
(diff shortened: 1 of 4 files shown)


src/chunk/cac.ts
```diff
@@ -41,20 +41,12 @@ export function makeContentAddressedChunk(payloadBytes: Uint8Array): Chunk {
...
-function makeChunk(data: ValidChunkData, address: () => ChunkAddress): Chunk {
-  const span = () => bytesAtOffset(CAC_SPAN_OFFSET, SPAN_SIZE, data)
-  const payload = () => flexBytesAtOffset(CAC_PAYLOAD_OFFSET, MIN_PAYLOAD_SIZE, MAX_PAYLOAD_SIZE, data)
 
   return {
     data,
-    span,
-    payload,
-    address,
+    span: () => span,
```
vojtechsimetka, line 47: Where did the bytesAtOffset method call disappear?
```ts
const span = () => bytesAtOffset(CAC_SPAN_OFFSET, SPAN_SIZE, data)
```

vojtechsimetka (commented):
LGTM but I'm not too familiar with SOC and CAC so I'll let somebody else to approve.

src/chunk/cac.ts
agazso: With the version before the changes, in the content addressed chunk only the chunk data was stored (as `data`) and everything else was calculated from that. In the new version the `span` is stored individually (captured in the scope of the returned object to be precise) and not calculated from the data.

src/chunk/cac.ts
```diff
@@ -71,16 +63,15 @@ export function isValidChunkData(data: Uint8Array, chunkAddress: ChunkAddress):
...
  *
  * @param data          The chunk data
  * @param chunkAddress  The address of the chunk
  *
  * @returns a valid content addressed chunk or throws error
  */
-export function verifyChunk(data: Uint8Array, chunkAddress: ChunkAddress): Chunk {
-  if (isValidChunkData(data, chunkAddress)) {
-    return makeChunk(data, () => chunkAddress)
+export function assertValidChunkData(data: Uint8Array, chunkAddress: ChunkAddress): void {
```
agazso, line 73: This does not seem to be an [assertion function](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) because it return type is not an assertion signature, therefore it does not provide type information to the Typescript compiler.

It would be better to change this in one of the following ways:
- change the return type to an assertion signature
- rename it to something else (e.g. `validateChunkData`)

agazso (approved):

PR #241, opened 2021-03-31
fix: correct package.json types path

AuHau:


vojtechsimetka (approved):

agazso (approved):

agazso (commented):
May I suggest to change the title of the PR to `fix: correct package.json types path` ?
It would be better understandable in the context of a changelog that is automatically generated.

PR #245, opened 2021-04-01
style: propagated linter rules that have been changed in swarm-cli and extension

vojtechsimetka:
@nugaon raised in the https://github.com/ethersphere/bee-status/pull/35 PR that the linter configuration is different in each project and that the swarm-cli and swarm-extension have newer rules resolving some eslint<->prettier clashes.

1. I tried to update the bee-js rules but now there actually are clashes. Can someone else try?
2. Once clashes are resolved and we agree on the rules here, I'll propagate them to other repos.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,9 +13,8 @@ module.exports = {
   },
   plugins: ['jest'],
   rules: {
-    'array-bracket-newline': ['error', { multiline: true }],
+    'array-bracket-newline': ['error', 'consistent'],
     strict: ['error', 'safe'],
-    curly: 'error',
     'block-scoped-var': 'error',
     complexity: 'warn',
     'default-case': 'error',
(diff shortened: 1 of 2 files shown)


nugaon (approved):

.eslintrc.js
```diff
@@ -73,7 +72,7 @@ module.exports = {
           requireLast: true,
         },
         singleline: {
-          delimiter: 'semi',
+          delimiter: 'comma',
```
nugaon, line 75: hmm.. it seems the old version is the good one then, and in the other projects it is wrong.
maybe the collusion is fixed here and hasn't been propagated to the other project's config.

nugaon: can you revert this line pls?

nugaon (approved):

.eslintrc.js
vojtechsimetka:
```suggestion
          delimiter: 'semi',
```

AuHau (approved):

PR #246, opened 2021-04-06
feat: high-level json feed api

### LINKED ISSUE #174: Feeds: make it easy to set json to feed
Hello! 

I have been playing around with feeds a bit. I'm having to do a lot of complicated and cumbersome things to just set a json value to a feed topic. I would like the following scenario to be possible:

const userData = { name: "michelle", address: "0x0", avatar: "some reference" }
const userFeed = await bee.setFeed(topic "userData", privatekey, value userData)
const retrievedUserFeed = await bee.getFeed(topic "userData", address) 

Right now i have to go through so many functions and steps that it is not feasible to expect the median react dev to understand this imo.

AuHau:
This is draft PR as I want to hear your opinion about my approach. Once we settle on some approach I will finish the PR with test coverage.

I was thinking about how to approach this with consideration that we already have some feed components in place. I came up with an abstraction layer that I named `DataFeed` which supports easily setting and getting data to/from a feed. I am not sure about the naming, but since we have already "writer" and "reader" for feeds I was a bit lost and came up with this name. If you have other suggestions, shoot them!

Also introduced direct methods on `Bee` that matches similar API that Michelle proposed.

Closes #174


src/feed/type.ts
```diff
@@ -1,5 +1,6 @@
 const feedTypes = ['sequence', 'epoch'] as const
 export type FeedType = typeof feedTypes[number]
+export const DEFAULT_FEED_TYPE = 'sequence' as FeedType
```
nugaon, line 3: Do not use `as` keywords if it is not necesarry
```suggestion
export const DEFAULT_FEED_TYPE: FeedType = 'sequence'
```

src/feed/data.ts
```diff
@@ -0,0 +1,47 @@
...
+    const jsonString = JSON.stringify(data)
+
+    return new TextEncoder().encode(jsonString)
+  } catch (e) {
+    e.message = `DataFeed: ${e.message}`
+    throw e
+  }
+}
+
+function getData<T extends Record<string, unknown> | number | string>(bee: Bee, writer: FeedWriter): () => Promise<T> {
```
nugaon, line 17: `writer` should be `reader` with type `FeedReader` here

src/feed/data.ts
```diff
@@ -0,0 +1,47 @@
...
+    const retrievedData = await bee.downloadData(feedUpdate.reference)
+
+    return JSON.parse(new TextDecoder().decode(retrievedData))
+  }
+}
+
+function setData<T extends Record<string, unknown> | number | string>(
+  bee: Bee,
+  writer: FeedWriter,
+): (data: T) => Promise<ReferenceResponse> {
```
nugaon, line 29: the generic is not justified here, moreover, it just makes harder the readability. Use `Record<string, unknown> | number | string` instead.

nugaon (commented):
I have some suggestions for the naming:

- getData -> getJsonData
- setData -> setJsonData
- *DataFeed* -> *JsonDataFeed*
- serializeData -> serializeJsonData
- setFeed -> setJsonFeed
- getFeed -> getJsonFeed

nugaon:
I just want to mention again my proposal which serves this use-case in an optimized way: https://github.com/ethersphere/bee/issues/1311

AuHau:
I am not sure if it is needed to put `JSON` in all the naming. Most people think about JSON as JS objects, while JSON.parse/stringify is able to handle most of other data types as well like simple strings, numbers, arrays etc.

src/feed/data.ts
```diff
@@ -0,0 +1,47 @@
+import { DataFeed, FeedWriter, ReferenceResponse } from '../types'
+import { Bee } from '../bee'
+
+function serializeData(data: Record<string, unknown> | number | string): Uint8Array {
+  if (typeof data === 'string') return new TextEncoder().encode(data)
```
agazso, line 5: I am not sure I understand how this is supposed to work in relation with `setData` and `getData`. If you pass in a string then it will be uploaded as binary data with `setData`, but in `getData` it will try to interpret it as JSON and it will throw an exception.

The easiest would be to always use JSON, even if a string is passed in. It is also possible to differentiate between strings and other types, but then I think it would be good to introduce a `deserialize` function which tries to parse the data as JSON and if it fails then tries to convert to text. In that case it would also make sense to deal with binary data (think of client-side encryption), but then the complexity (and the number of potential edge-cases) starts to be so high that it's questionable to support it in this interface.

src/bee.ts
```diff
@@ -468,6 +472,57 @@ export class Bee {
     return makeFeedWriter(this.url, type, canonicalTopic, canonicalSigner)
   }
 
+  /**
+   * Makes a DataFeed which allows easily setting and getting of data to feeds.
+   * Especially handy if you manipulate feed's data more then once.
+   *
+   * @param writer
+   */
+  makeDataFeed<T extends Record<string, unknown> | number | string>(writer: FeedWriter): DataFeed<T> {
```
agazso, line 481: If we expose this in the Bee class there will be three ways to write to a feed:
1. using `makeFeedWriter`
2. using `setFeed`
3. using `makeDataFeed`

The justification for the 1st is that it is the low-level interface that gives you access to the feed API directly. The 2nd is the convenience interface when you just want to store JSON data. The 3rd one seems to me an optimisation of the 2nd case when you are doing many writes.

If that's the case I found that this puts a bit of a too high mental load on the user, because they will need to understand the three cases to understand which fits their use-case best. That also puts the burden on us to explain and communicate the differences in our documentation. Maybe it would be better to not expose the 3rd one to reduce this cognitive overhead. If someone wants to optimise their use-case it's always possible using the 1st option.

What do you think?

agazso:
I just wanted to mention that we also have our `Data` type that supports conversion to JSON, text and hex. Did you consider using that or something similar? I think it would give consistency to our API if we reused the same types in similar use-cases.

src/feed/data.ts
AuHau: Yeah, you are right. I wrote this before I did some verification on how JSON.parse/stringify works and string should not be left out as a special case but should pass through JSON.stringify as well.

nugaon:
> I am not sure if it is needed to put Json in all the naming. Most people think about JSON as JS objects, while JSON.parse/stringify is able to handle most of other data types as well like simple strings, numbers, null, arrays etc. which I am to support as well.

Well, I think `Json` is JSON regardless what devs think and the types also notate this functionality. Don't you think it describes itself and make clearer in functionality? Because you cannot pass any `Data` in that sense like Uint8Array.

src/bee.ts
AuHau: >The justification for the 1st is that it is the low-level interface that gives you access to the feed API directly. The 2nd is the convenience interface when you just want to store JSON data. The 3rd one seems to me an optimisation of the 2nd case when you are doing many writes.

Yeah, you got that right.

I guess you have a valid point with the mental load, but still, I feel it is a justified use-case. What about if I would expose the `makeDataFeed` as a utility and not on the `Bee` class itselfM

AuHau:
@agazso good point about the `Data`. I did not really think about that. Hmmm I am not sure if having that extra layer to access the data is desired though (the extra `.json()` call).

I guess this brings two possible paths on how to move further with this PR.

 1. Use `Data` interface as we use in other APIs. In that case, I would keep the naming as is now.
 2. Go with JSON-only content. In that case, I would refactor the names based on @nugaon suggestion.

What do you think?

IxaBrjnko:
I really appreciate all of the hard work going into this, it is inspiring!  I am wrapping up with quarter end, and will be diving in deep to feeds in bee-js soon... my particular use case involves uint8 array encoding of video data, and while I will definitely need json functionality as well, I wonder if forcing all data types through json conversion is necessary for your application of this module.  However, as long as there remains a method for editing feeds that are not json formatted somewhere in bee-js, I should be just fine....

agazso:
> I really appreciate all of the hard work going into this, it is inspiring! I am wrapping up with quarter end, and will be diving in deep to feeds in bee-js soon... my particular use case involves uint8 array encoding of video data, and while I will definitely need json functionality as well, I wonder if forcing all data types through json conversion is necessary for your application of this module. However, as long as there remains a method for editing feeds that are not json formatted somewhere in bee-js, I should be just fine....

Hey @IxaBrjnko, thanks for the kind words!

It will be always possible to work with binary data with the low-level `FeedReader`/`FeedWriter` API.  For example a similar implementation for working with `Uint8Array` would look like this:

```typescript
// value is of type Uint8Array
export const setUint8ArrayFeed = async (topic, value, pk) => {
    try {
        const encodedTopic = await bee.makeFeedTopic(topic)
        const swarmReference = await bee.uploadData(value)
        const feedWriter = bee.makeFeedWriter('sequence', encodedTopic, pk)
        const response = await feedWriter.upload(swarmReference)
        return response
    } catch (error) {
        console.error(error)
    }
}
export const getUint8ArrayFeed = async (topic, address) => {
    try {
        const encodedTopic = await bee.makeFeedTopic(topic)
        const feedReader = bee.makeFeedReader('sequence', encodedTopic, address)
        const feedUpdate = await feedReader.download()
        const data = bee.downloadData(feedUpdate.reference)
        return data
    } catch (error) {
        console.error(error)
    }

}
```

src/bee.ts
agazso: My gut instinct says [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) and also that's it easy to add when someone specifically requests this feature. 

I would rather add to the documentation of `FeedReader` and `FeedWriter` how you can implement your own 'data feeds' for different datatypes (other than JSON), like this: https://github.com/ethersphere/bee-js/pull/246#issuecomment-814190538

AuHau: Sounds good!

AuHau:
So I have taken the path of JSON-only adoption for this PR so I have refactored the naming as @nugaon suggested.

src/types/index.ts
```diff
@@ -148,6 +148,22 @@ export interface FeedReader {
...
+ * data from feeds. It works closely with JSON.parse/stringify so all supported data
+ * types by that is also supported by this abstraction.
+ */
+export interface JsonFeed<T extends AnyJson> {
+  readonly writer: FeedWriter
+  set(data: AnyJson): Promise<ReferenceResponse>
+  get(): Promise<T>
+}
+
+export interface DataFeedOptions {
```
agazso, line 162: As this is exposed to the public API as well and there are no other mentioning of DataFeeds, wouldn't this be better called `JsonFeedOptions`?

src/feed/json.ts
```diff
@@ -0,0 +1,39 @@
+import { JsonFeed, FeedWriter, ReferenceResponse, FeedReader, AnyJson } from '../types'
+import { Bee } from '../bee'
+
+function serializeJson(data: AnyJson): Uint8Array {
+  try {
+    const jsonString = JSON.stringify(data)
+
+    return new TextEncoder().encode(jsonString)
+  } catch (e) {
+    e.message = `DataFeed: ${e.message}`
```
agazso, line 10: Isn't this `JsonFeed`?

src/types/index.ts
AuHau: Thanks, I missed this in the refactor!

src/feed/json.ts
AuHau: Thanks, I missed this in the refactor!

agazso (approved):

src/bee.ts
```diff
@@ -468,6 +472,43 @@ export class Bee {
     return makeFeedWriter(this.url, type, canonicalTopic, canonicalSigner)
   }
 
+  /**
+   * High-level function that allows you to easily set data to feed.
```
nugaon, line 476:
```suggestion
   * High-level function that allows you to easily set JSON data to feed.
```

nugaon (approved):
LGTM!

PR #255, opened 2021-04-09
refactor!: verify to assert

### LINKED ISSUE #211: Use `assert...` instead of `verify` functions

AuHau:
Closes #211 

This is work I had in progress already since last week, so I have wrapped it up as it was getting rather stale.

I have refactored the `verify*` functions into `assert*` functions. Since this is already breaking change for example thanks to the `verifyBytes()` refactor, I took the liberty and also changed the order of the parameters in some of the functions as I believe that the order should follow two guidelines:

 1. Order of significants
 2. Modifiers should come after the subject. E.g. if we are asserting bytes length, this is basically saying "assert that bytes under variable `data` (first parameter) should have length `20` (second parameter)" instead of the current approach that follows more "assert that length `20` (first parameter) is matching variable `data` (second parameter)

### DIFF
--- a/src/utils/bytes.ts
+++ b/src/utils/bytes.ts
@@ -31,7 +31,7 @@ export interface FlexBytes<Min extends number, Max extends number> extends Uint8
  * @param length  The length of the byte array
  * @param b       The byte array
  */
-export function isBytes<Length extends number>(length: Length, b: Uint8Array): b is Bytes<Length> {
+export function isBytes<Length extends number>(b: unknown, length: Length): b is Bytes<Length> {
   return b instanceof Uint8Array && b.length === length
 }
 
@@ -41,11 +41,10 @@ export function isBytes<Length extends number>(length: Length, b: Uint8Array): b
  * @param length  The specified length
  * @param b       The byte array
  */
-export function verifyBytes<Length extends number>(length: Length, b: Uint8Array): Bytes<Length> | never {
-  if (isBytes(length, b)) {
-    return b
+export function assertBytes<Length extends number>(b: unknown, length: Length): asserts b is Bytes<Length> {
+  if (!isBytes(b, length)) {
+    throw new TypeError(`Parameter is not valid Bytes of length: ${length} !== ${(b as Uint8Array).length}`)
   }
-  throw new Error(`verifyBytes failed, length: ${length} !== ${b.length}`)
 }
 
 /**
@@ -58,9 +57,9 @@ export function verifyBytes<Length extends number>(length: Length, b: Uint8Array
 export function isFlexBytes<Min extends number, Max extends number = Min>(
   min: Min,
   max: Max,
-  b: Uint8Array,
+  b: unknown,
 ): b is FlexBytes<Min, Max> {
-  return b.length >= min && b.length <= max
+  return b instanceof Uint8Array && b.length >= min && b.length <= max
 }
 
 /**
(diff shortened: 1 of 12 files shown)


vojtechsimetka (approved):
LGTM

src/utils/bytes.ts
```diff
@@ -58,9 +57,9 @@ export function verifyBytes<Length extends number>(length: Length, b: Uint8Array
 export function isFlexBytes<Min extends number, Max extends number = Min>(
   min: Min,
   max: Max,
-  b: Uint8Array,
+  b: unknown,
```
nugaon, line 60: Here you don't follow your convention (as you can see below at `assertFlexBytes`)

src/utils/bytes.ts
```diff
@@ -70,15 +69,16 @@ export function isFlexBytes<Min extends number, Max extends number = Min>(
  * @param max     Maximum size of the array
  * @param b       The byte array
  */
-export function verifyFlexBytes<Min extends number, Max extends number = Min>(
+export function assertFlexBytes<Min extends number, Max extends number = Min>(
```
nugaon, line 72: jsdoc parameter order is wrong

src/utils/bytes.ts
```diff
@@ -88,8 +88,13 @@ export function verifyFlexBytes<Min extends number, Max extends number = Min>(
  * @param length The length of data to be returned
  * @param data   The original data
  */
-export function bytesAtOffset<Length extends number>(offset: number, length: Length, data: Uint8Array): Bytes<Length> {
-  return data.slice(offset, offset + length) as Bytes<Length>
+export function bytesAtOffset<Length extends number>(data: Uint8Array, offset: number, length: Length): Bytes<Length> {
```
nugaon, line 91: jsdoc parameter order is wrong

src/utils/bytes.ts
```diff
@@ -31,7 +31,7 @@ export interface FlexBytes<Min extends number, Max extends number> extends Uint8
  * @param length  The length of the byte array
  * @param b       The byte array
  */
-export function isBytes<Length extends number>(length: Length, b: Uint8Array): b is Bytes<Length> {
+export function isBytes<Length extends number>(b: unknown, length: Length): b is Bytes<Length> {
```
nugaon, line 34: jsdoc parameter order is wrong

src/utils/bytes.ts
```diff
@@ -41,11 +41,10 @@ export function isBytes<Length extends number>(length: Length, b: Uint8Array): b
  * @param length  The specified length
  * @param b       The byte array
  */
-export function verifyBytes<Length extends number>(length: Length, b: Uint8Array): Bytes<Length> | never {
-  if (isBytes(length, b)) {
-    return b
+export function assertBytes<Length extends number>(b: unknown, length: Length): asserts b is Bytes<Length> {
```
nugaon, line 44: jsdoc parameter order is wrong

nugaon (changes requested):
I also like it in this new structure, but there were some JSDoc issues, can you fix those?

AuHau:
Thanks @nugaon will fix those! I thought that if the name of variable is mentioned the order does not matter, but I guess it matters.

nugaon (approved):

agazso (approved):
Looks good!

PR #258, opened 2021-04-12
refactor: split integration and unit tests

AuHau:
So in the end I managed to work around the ordering of tests using Jest's custom sequencer (`./test/test-type-sequencer.js`). It is in JS as Jest does not support TS for this.

So the current NPM scripts layout is:

 - `npm test` - run all tests in one Jest execution ordered by unit tests first
 - `npm run test:unit` - run all unit tests (currently only for node)
 - `npm run test:integration` - run integration tests for both node and browser (also compile the sources)
 - `npm run test:integration:browser|node` - run integration tests for given environment
 - `npm run test:node|browesr` - run tests for given environment. Tests are ordered by unit tests first.

The set of the scripts (shortcuts) can be cut down if you think some of those are not necessary.


.github/workflows/tests.yaml
```diff
@@ -74,7 +74,7 @@ jobs:
         if: steps.cache-npm.outputs.cache-hit != 'true'
         run: npm ci
 
-      - name: Run nodejs tests
+      - name: Run unit tests
```
nugaon, line 77: I don't get it fully because you wrote 
> npm run test:node|browesr` - run tests for given environment. Tests are ordered by unit tests first.

So therefore it runs unit tests and then integration tests or doesn't it?

```suggestion
      - name: Run unit and integration tests
```

package.json
```diff
@@ -34,9 +34,13 @@
     "compile:types": "tsc --emitDeclarationOnly --declaration",
     "compile:browser": "webpack --progress --env target=web",
     "docs": "rimraf docs && typedoc",
-    "test": "npm run test:node && npm run test:browser",
-    "test:browser": "npm run compile:browser && jest --verbose --config=jest.config.ts --selectProjects=dom",
-    "test:node": "jest --verbose --config=jest.config.ts --selectProjects=node",
+    "test": "npm run compile:browser && jest --verbose --config=jest.config.ts --selectProjects=node:unit node:integration dom:integration",
+    "test:integration:browser": "npm run compile:browser && jest --verbose --config=jest.config.ts --selectProjects=dom:integration",
+    "test:integration:node": "jest --verbose --config=jest.config.ts --selectProjects=node:integration",
+    "test:integration": "npm run compile:browser && jest --verbose --config=jest.config.ts --selectProjects=node:integration dom:integration",
```
nugaon, line 40: Later we would change one of the flag of browser or node integration whithin its defined script, but we may forget anytime to do that here. this suggestion sort out this problem. 

```suggestion
    "test:integration": "npm run compile:browser && npm run test:integration:browser && npm run test:integration:node",
```

package.json
```diff
@@ -34,9 +34,13 @@
     "compile:types": "tsc --emitDeclarationOnly --declaration",
     "compile:browser": "webpack --progress --env target=web",
     "docs": "rimraf docs && typedoc",
-    "test": "npm run test:node && npm run test:browser",
-    "test:browser": "npm run compile:browser && jest --verbose --config=jest.config.ts --selectProjects=dom",
-    "test:node": "jest --verbose --config=jest.config.ts --selectProjects=node",
+    "test": "npm run compile:browser && jest --verbose --config=jest.config.ts --selectProjects=node:unit node:integration dom:integration",
```
nugaon, line 37: I think it is more straightforward with less error possibility.

```suggestion
    "test": "npm run test:unit && npm run test:integration",
```

nugaon (commented):
I think this solution is little bit overcomplicated, because `jest` supports `projects` by default (that you use correctly anyway) and you could manage this with much simplier script commands if you just use only that (complexity always leads to higher error propability). What was your consideration to introduce `test-type-sequencer.js`?

AuHau:
@nugaon I am not sure what complexity you are pointing at? Do you mean there are too many npm-scripts? As I have pointed out in the PR's description I am fine with cutting it down, but I am not sure what are you proposing?

Maybe if I understand it right, are you proposing to use the Jest projects instead of the npm-scripts shortcuts? So basically there would be only `test` script with the definition `npm run compile:browser && jest --verbose --config=jest.config.ts` and if you want to run only some specific then you would have to run `npm test -- --selectProjects=node:unit node:integration` for example?

> What was your consideration to introduce test-type-sequencer.js?

I touched that in the PR's description, but to expand. One of the goals of my proposal was to run unit tests first so they would work partially as smoke tests, so the tests have to be ordered somehow.
I found that Jest's [`testSequencer`](https://jestjs.io/docs/configuration#testsequencer-string) allows that, so I wrote a custom sequencer that orders tests in order first unit tests and then integration tests.

package.json
AuHau: I am against this. The whole idea behind using the custom `testSequencer` was so all tests would run in one process to speed up the execution, for example avoiding double test discovery etc. This would execute the tests in two processes.

package.json
AuHau: I am against this. See my [comment](https://github.com/ethersphere/bee-js/pull/258#discussion_r612974394) below.

.github/workflows/tests.yaml
AuHau: Thanks, good catch! I started to do some tweaks to the GitHub actions, but then changed the approach a bit so it was not needed, but forgot to revert it.

nugaon:
> @nugaon I am not sure what complexity you are pointing at? Do you mean there are too many npm-scripts? As I have pointed out in the PR's description I am fine with cutting it down, but I am not sure what are you proposing?

The count of the NPM scripts is totally fine, I just pointed out [in my comment](https://github.com/ethersphere/bee-js/pull/258#discussion_r612401353) if you want to change the parameters of one particular test script, you have to change on every related command which would invoke this test type. E.g. you change options of `test:integration:browser` because of some reason (for instance pass process env variable), then you have to change it any related commands, namely `test` and `test:integration` as well (this is also related the compiling for browser test and any other dependencies).
On the other hand if these group commands like `test:integration` calls the corresponding npm test processes instead of test projects, you don't have to define these before mentioned additional parameters, actions for the test and the script itself is also simplier.

> I touched that in the PR's description, but to expand. One of the goals of my proposal was to run unit tests first so they would work partially as smoke tests, so the tests have to be ordered somehow.

The test can be ordered in that way as well that I proposed in the review suggestion.

> The whole idea behind using the custom testSequencer was so all tests would run in one process to speed up the execution, for example avoiding double test discovery etc. This would execute the tests in two processes.

I think the decision here is either decrease the complexity level and thereby the error probability or have quicker tests. I don't know how much faster the latter than the former, but what is your opinion about this with this plus info of my proposal?

AuHau:
Oh, I understand your point now. Yeah, you are right about that. Hmmm. 

Honestly, I don't have real data to back up the speed up of the tests, also because of the external dependencies of integrations tests, I think that some simple benchmark would not yield any trustworthy data. And I don't think some more advanced benchmarking is not worth the time.

In my experience, you usually set up these things one time and then don't really change it, so I would argue for the (potential) speedup of tests that we run every day over some potential change in (far) future.

nugaon (approved):
Let's be then, but we have to keep in mind this, anyway the code seems work well.

AuHau:
PS.: @nugaon please check Mattermost ;-)

PR #259, opened 2021-04-14
fix: use bigint for bzz amounts

### LINKED ISSUE #254: Return all balances, settlements and money value as `BigInt`
Currently all the balance, settlement, chequebook values are being returned in the API as just numbers. This is not going to work for anything above `0.9 BZZ`  which is completely realistic value to get. The API returns it as a number JSON object which does not restrict precision, but it needs to be parsed in a right way using for example [`json-bigint`](https://www.npmjs.com/package/json-bigint). 

Here is an article which could be a good start [Handling BigInt in Axios using json-big](https://itstalwar15.medium.com/handling-bigint-in-axios-using-json-big-in-javascript-d915ae85ffc0)

AuHau:
So I have wrapped the BigInt work. I have introduced a custom `forceBigInt` config flag for the `safeAxios` call that runs the parsing with `JsonBigNative.parse()` in a way that all numbers are converted to BigInt no matter if they exceeded the threshold or not. This is important to have a consistent API. If this flag is not specified then only values exceeding the threshold are converted to `BigInt`.

Closes #254.

### DIFF
--- a/src/utils/safeAxios.ts
+++ b/src/utils/safeAxios.ts
@@ -6,6 +6,13 @@ import utils from 'axios/lib/utils'
 import normalizeHeaderName from 'axios/lib/helpers/normalizeHeaderName'
 axios.defaults.adapter = require('axios/lib/adapters/http') // https://stackoverflow.com/a/57320262
 
+/**
+ * Utility function from axios's implementation
+ * https://github.com/axios/axios/blob/d99d5faac29899eba68ce671e6b3cbc9832e9ad8/lib/defaults.js
+ *
+ * @param headers
+ * @param value
+ */
 function setContentTypeIfUnset(headers: Record<string, unknown>, value: string) {
   if (!utils.isUndefined(headers) && utils.isUndefined(headers['Content-Type'])) {
     headers['Content-Type'] = value
--- a/src/utils/safeAxios.ts
+++ b/src/utils/safeAxios.ts
@@ -1,13 +1,86 @@
 import axios, { AxiosRequestConfig, AxiosResponse } from 'axios'
 import { BeeError, BeeRequestError, BeeResponseError } from './error'
+import JsonBig from 'json-bigint'
 
+import utils from 'axios/lib/utils'
+import normalizeHeaderName from 'axios/lib/helpers/normalizeHeaderName'
 axios.defaults.adapter = require('axios/lib/adapters/http') // https://stackoverflow.com/a/57320262
 
-export async function safeAxios<T>(config: AxiosRequestConfig): Promise<AxiosResponse<T>> {
+function setContentTypeIfUnset(headers: Record<string, unknown>, value: string) {
+  if (!utils.isUndefined(headers) && utils.isUndefined(headers['Content-Type'])) {
+    headers['Content-Type'] = value
+  }
+}
+
+/**
+ * Taken over from axios's implementation
+ * https://github.com/axios/axios/blob/d99d5faac29899eba68ce671e6b3cbc9832e9ad8/lib/defaults.js
+ *
+ * Unfortunately we can't use the default implementation as fallback one and override only the JSON serialization as the JSON.stringify
+ * is a last thing to do as all the previous cases (buffer, file etc.) are objects as well
+ * so they have to be ruled out first before trying JSON.stringify.
+ *
+ * @param data
+ * @param headers
+ */
+function jsonBigTransformRequest(data: unknown, headers: Record<string, unknown>): string | unknown {
+  normalizeHeaderName(headers, 'Accept')
+  normalizeHeaderName(headers, 'Content-Type')
+
+  if (
+    utils.isFormData(data) ||
+    utils.isArrayBuffer(data) ||
+    utils.isBuffer(data) ||
+    utils.isStream(data) ||
+    utils.isFile(data) ||
+    utils.isBlob(data)
+  ) {
+    return data
+  }
+
+  if (utils.isArrayBufferView(data)) {
+    return data.buffer
+  }
+
+  if (utils.isURLSearchParams(data)) {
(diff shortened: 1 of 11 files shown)


src/utils/safeAxios.ts
```diff
@@ -1,13 +1,86 @@
...
+function jsonBigTransformResponse(forceBigInt = false) {
+  return (data: unknown): unknown => {
+    if (typeof data === 'string') {
+      const JsonBigNative = JsonBig({ useNativeBigInt: true, alwaysParseAsBig: forceBigInt })
+
+      try {
+        data = JsonBigNative.parse(data)
+      } catch (e) {
+        /* Ignore */
+      }
```
agazso, line 70: Instead of overwriting data I think it's cleaner to indicate in the control flow what happens.
```suggestion
      try {
        return JsonBigNative.parse(data)
      } catch (e) {
        return data
      }
```

src/utils/safeAxios.ts
```diff
@@ -1,13 +1,86 @@
...
+    return JsonBigNative.stringify(data)
+  }
+
+  return data
+}
+
+function jsonBigTransformResponse(forceBigInt = false) {
+  return (data: unknown): unknown => {
+    if (typeof data === 'string') {
+      const JsonBigNative = JsonBig({ useNativeBigInt: true, alwaysParseAsBig: forceBigInt })
```
agazso, line 64: It seems that the `json-bigint` package has a dependency of the `bignumber.js` package and when using `useNativeBigInt: false` it uses that instead of the native `BigInt` support from the browser.

My concern is that `BigInt` is still not very well supported, see an earlier discussion regarding this:
https://github.com/ethersphere/bee-js/pull/207#discussion_r597578613

I wonder how hard would be to detect `BigInt` support in the browser and set the `useNativeBigInt` flag based on the result of that check?

AuHau: Yeah, you are right about `useNativeBigInt`.

Honestly, I think that `BigInt` support is good enough for our use case. I mean we are developing new edge-cutting technology, so I would not expect to support too old or browsers (like IE) or some edge-case browsers. 

Also, I would be against this un-consistent behavior as these two approaches are not interchangeable. We either should fully switch to `bignumber.js` or keep `BigInt` but with having both the devs would have to test-case for each possibility...

vojtechsimetka: I would also second just using `BigInt` across the board. `BigNumber` could be ambiguous when it comes to the result - is it decimal number or integer? `BigInt` makes this clear.

When we get a report of users needing support for older/minority browsers we can deal with it then. I am pretty sure we would have to sort out other issues as well not just this one.

agazso: Maybe I'm sensitive to this topic because currently I'm using an older mac and iPhone that runs Safari 13.1 and that does not support it 😄 

As I wrote earlier I think the problem is that people will make applications on top of `bee-js` and they might not know about this limitation so they might be surprised and won't be able to mitigate it.

On the other hand I feel that by putting `const useNativeBigInt = typeof BigInt === 'function'` and use that variable would provide a simple solution, because for practical use-cases `json-bigint` use `bignumber.js` to polyfill BigInt as it used it originally before the browser `BigInt` support was added to it. So considering we trust `json-bigint` this is supposed to work as well.

It's true that it makes it harder to test, but as far as I understand we also don't have browser test for this either. Anyways maybe it's better just to open an issue about it right now and decide to work on it later.

src/utils/safeAxios.ts
```diff
@@ -1,13 +1,86 @@
 import axios, { AxiosRequestConfig, AxiosResponse } from 'axios'
 import { BeeError, BeeRequestError, BeeResponseError } from './error'
+import JsonBig from 'json-bigint'
 
+import utils from 'axios/lib/utils'
+import normalizeHeaderName from 'axios/lib/helpers/normalizeHeaderName'
 axios.defaults.adapter = require('axios/lib/adapters/http') // https://stackoverflow.com/a/57320262
 
-export async function safeAxios<T>(config: AxiosRequestConfig): Promise<AxiosResponse<T>> {
+function setContentTypeIfUnset(headers: Record<string, unknown>, value: string) {
```
agazso, line 9: This is also something taken from axios' implementation?

agazso (approved):

src/utils/safeAxios.ts
vojtechsimetka: I think this indeed is a separate issue. We already use `BigInt` elsewhere so we already don't support the older browsers.

IMHO the first step would be to actually determine what browsers we currently support and which we do not.

vojtechsimetka (approved):

src/utils/safeAxios.ts
AuHau: I agree with Vojtech here.

Hmm, @agazso are you sure `bignumber.js` polyfills `BigInt` type? I honestly don't think so as `BigInt` is very much an internal thing as it creates a whole new type in the JS type system (eq. `typeof someBigInt === 'bigint'`), supports arithmetic operators etc and I don't think that is possible to polyfill this in JS. I mean yes `bignumber.js` provides its own API that does the same job as `BigInt`, but it does not polyfill it and that is my problem. We would have to have return types of these methods basically `BigInt | BigNumber` which is something I am against as every time the users of this library would have called these methods they would have to type check what object they actually got returned.

src/utils/safeAxios.ts
AuHau: Yes

src/utils/safeAxios.ts
agazso: Ok, I see now. In that case I agree it's better to just use `BigInt`.

src/utils/safeAxios.ts
agazso: In that case it would be good to add a comment similarly to `jsonBigTransformRequest`

nugaon (approved):

Issue #156, opened 2021-02-26
Missing tests for debug chunk endpoints


agazso:
this function is not used anywhere currently.

_Originally posted by @nugaon in https://github.com/ethersphere/bee-js/pull/151#discussion_r583615991_

PR #262, opened 2021-04-15
feat: expose pinning functions

Cafe137:


agazso (commented):
It would be good to have tests to at least try the happy path and verify that the returned data is what we expect.

See the other pinning functions for reference.

AuHau:
Generally LGTM, it would be good to include tests as well, but if it is important to have it in the release I guess we can do that as part of #156 work.

AuHau (approved):
LGTM!

PR #264, opened 2021-04-15
fix: check typings for test folder

### LINKED ISSUE #257: Typings are not verified for tests
I believe that typings are not checked for our tests. I think this mainly comes from the fact that Jest uses webpack for the compilation of the TS files and webpack does not do any verifications as `tsc` does.

We could create another check inside our GitHub workflow that would check this. Or other suggestions are welcomed.

AuHau:
Creates new npm script `check:types` that runs types check also for test folder. Unfortunately, I had to create a new `tsconfig` file that extends our original one and adds the `test` folder.

I have also removed building of the source code for node and browser from GitHub Workflow because I believe it does not detect anything meaningful and only takes unnecessary time.

Closes #257

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -50,14 +50,8 @@ jobs:
       - name: Dependency check
         run: npm run depcheck
 
-      - name: Build nodejs
-        run: npm run compile:node -- --env mode=production
-
-      - name: Build browser
-        run: npm run compile:browser -- --env mode=production
-
-      - name: Build typings
-        run: npm run compile:types
+      - name: Check typings
+        run: npm run check:types
 
       - name: Build docs
         run: npm run docs
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "test:unit": "jest --verbose --config=jest.config.ts --selectProjects=node:unit",
     "test:node": "jest --verbose --config=jest.config.ts --selectProjects=node:unit node:integration",
     "test:browser": "npm run test:integration:browser",
-    "check:types": "tsc --project tsconfig.test.json --noEmit",
+    "check:types": "tsc --project tsconfig.test.json",
     "lint": "eslint --fix \"src/**/*.ts\" \"test/**/*.ts\" && prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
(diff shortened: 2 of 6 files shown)


.github/workflows/check.yaml
```diff
@@ -47,14 +47,8 @@ jobs:
         env:
           CI: true
 
-      - name: Build nodejs
-        run: npm run compile:node -- --env mode=production
-
-      - name: Build browser
-        run: npm run compile:browser -- --env mode=production
```
vojtechsimetka, line 54: I am fine with this being removed but I am sure it ended up here for a reason (I remember adding it). Maybe somebody else can remember.

Btw maybe we can consider running the `prepare` script here to make sure everything compiles fine for the npmjs release.

vojtechsimetka (approved):

.github/workflows/check.yaml
AuHau: Having `prepare` would be the same as having these steps, so I don't think it is desired to have it.

nugaon: As I remember the production build pointed out some error once, but since the `prepare` script runs before tests and npm publish as well I don't think it is necessary here.

package.json
```diff
@@ -41,6 +41,7 @@
     "test:unit": "jest --verbose --config=jest.config.ts --selectProjects=node:unit",
     "test:node": "jest --verbose --config=jest.config.ts --selectProjects=node:unit node:integration",
     "test:browser": "npm run test:integration:browser",
+    "check:types": "tsc --project tsconfig.test.json --noEmit",
```
nugaon, line 44: `--noEmit` should be placed in the referred tsconfig.

nugaon (approved):
please check out my comment for `--noEmit`

.github/workflows/check.yaml
vojtechsimetka: Yeah but if the dependencies are cashed, the `prepare` script does not run. Lets see if we ever get into trouble, I am fine leaving it out :)

nugaon: hmm if it does not run if there are no new deps then the `publish_npmjs.yaml` workflow is not good as it is.

AuHau: There it is not an issue as:

> prepare 
>   - Runs any time before the package is packed, i.e. during npm publish and npm pack

https://docs.npmjs.com/cli/v7/using-npm/scripts

PR #267, opened 2021-04-17
test: fixes missed escape character in `test/bee.sh`

significance:
just a simple fix so i can through the process, nice work guys ;D


vojtechsimetka (approved):

nugaon (approved):
thanks for the fix!

PR #268, opened 2021-04-17
chore: add check for unused or missing dependencies

vojtechsimetka:
We talked about this in past and since I added this to my personal project I figured I might as well create PR here. Let me know what do you think :)


AuHau (approved):
Sweet! Like it! :-) Maybe I would suggest ignoring `devDependencies`? But not sure if it is possible to do with the tool.

vojtechsimetka:
@AuHau yeah it is possible but I would personally not go that route. Even now we were able to remove some 4 unused `devDependencies`.

nugaon (approved):
I also like it, lincense is also OK.

PR #271, opened 2021-04-19
feat: enable setting default headers

### LINKED ISSUE #251: Append Extra Headers for All BeeJS requests
If there is a `proxy` between the dApp and the Bee client, the requests would require additional headers.

At Bee class initialization, there should be an optional `extraHeaders` parameter, where user can define these specific headers for their needs. It has to apply on all outgoing request from Bee instance.

AuHau:
Closes #251.

Not sure if or how to test this. Should this have some more in-depth documentation on `bee-js-docs`?


vojtechsimetka:
Can this be tested with some mock backend? Does it make sense?

nugaon (changes requested):
I Googled how to apply a header to all requests of axios and I saw `axios.defaults.headers.common` everywhere. (Axios doc: https://github.com/axios/axios#global-axios-defaults)

Have you tried out manually whether it works or not?

AuHau:
Damn, you are right 🙈 Sorry about that. Ok, I will test it then 😂

@vojtechsimetka yeah, will do that.

AuHau:
Heh, so even it is not documented my approach actually worked @nugaon 😂 But yeah that was pure luck as I was looking on a different section of docs when I wrote that.

Anyway, I have added a test for this, but had to add new dev dependency for that.

nugaon (approved):

PR #274, opened 2021-04-20
chore: define release checklist

AuHau:
This creates release checklist that is passed into release PRs by [bee-runner](https://github.com/ethersphere/bee-runner/pull/5), that we defined as our release process.

The communication points are left out here as it is Attilas responsibility.


agazso (approved):

agazso (approved):
👍

PR #275, opened 2021-04-20
test: unify integration setup

### LINKED ISSUE #270: Revisit integration testing setup
This morning @Cafe137 asked me for a help with integration tests as they were failing with `ECONNREFUSED`.

I am not sure how the integration testing setup was created. Currently the local bee nodes are by default searched at `http://bee-0.localhost`. I assume this is thanks to [`bee-local`](https://github.com/ethersphere/bee-local)? But at the same time we have the `./test/bee.sh` script that orchestrates the cluster with docker and is presented as the way to go to run the test but the nodes are not found under this URL which makes the test fails. It is possible to configure that with env. variable, but I think this should work out-of-the-box. 

So I would either change the defaults from `http://bee-0.localhost` to `http://localhost:1633` or fix the `./test/bee.sh` script to be reachable under that address.

@nugaon @agazso what do you think?

AuHau:
Closes #270

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -14,8 +14,8 @@ env:
   REPLICA: 5
   BEE_API_URL: 'http://bee-0.localhost'
   BEE_PEER_API_URL: 'http://bee-1.localhost'
-  BEE_DEBUG_API_URL: 'http://bee-0.localhost:1635'
-  BEE_PEER_DEBUG_API_URL: 'http://bee-1.localhost:1635'
+  BEE_DEBUG_API_URL: 'http://bee-0-debug.localhost'
+  BEE_PEER_DEBUG_API_URL: 'http://bee-1-debug.localhost'
   BEE_TEST_CHEQUEBOOK: true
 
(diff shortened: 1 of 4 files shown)


nugaon (approved):
LGTM

.github/workflows/tests.yaml
```diff
@@ -14,6 +14,8 @@ env:
   REPLICA: 5
   BEE_API_URL: 'http://bee-0.localhost'
   BEE_PEER_API_URL: 'http://bee-1.localhost'
+  BEE_DEBUG_API_URL: 'http://bee-0.localhost:1635'
```
agazso, line 17: Previously this was `http://bee-0-debug.localhost`

.github/workflows/tests.yaml
```diff
@@ -14,6 +14,8 @@ env:
...
+  BEE_PEER_DEBUG_API_URL: 'http://bee-1.localhost:1635'
```
agazso, line 18: Previously this was `http://bee-1-debug.localhost`

PR #276, opened 2021-04-20
fix: support address option for getJsonFeed

AuHau:
This address the shortcoming of design of `getJsonFeed` that required always to have the private key for fetching the data even if it is not required and basically it disallowed to fetch other's users feed data.

### DIFF
--- a/src/bee.ts
+++ b/src/bee.ts
@@ -41,11 +41,10 @@ import { downloadSingleOwnerChunk, uploadSingleOwnerChunkData } from './chunk/so
 import { makeTopic, makeTopicFromString } from './feed/topic'
 import { createFeedManifest } from './modules/feed'
 import { assertBeeUrl, stripLastSlash } from './utils/url'
-import { EthAddress, HexEthAddress, makeEthAddress, makeHexEthAddress } from './utils/eth'
+import { EthAddress, makeEthAddress, makeHexEthAddress } from './utils/eth'
 import { wrapBytesWithHelpers } from './utils/bytes'
 import { assertReference } from './utils/type'
 import { setJsonData, getJsonData } from './feed/json'
-import { add } from 'husky'
 
 /**
  * The Bee class provides a way of interacting with the Bee APIs based on the provided url
--- a/src/bee.ts
+++ b/src/bee.ts
@@ -548,29 +548,41 @@ export class Bee {
    * High-level function that allows you to easily get data from feed.
    * Returned data are parsed using JSON.parse().
    *
+   * This method also supports specification of `signer` object passed to constructor. The order of evaluation is:
+   *  - `options.address`
+   *  - `options.signer`
+   *  - `this.signer`
+   *
+   * At least one of these has to be specified!
+   *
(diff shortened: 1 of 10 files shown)


src/bee.ts
```diff
@@ -41,10 +41,11 @@ import { downloadSingleOwnerChunk, uploadSingleOwnerChunkData } from './chunk/so
 import { makeTopic, makeTopicFromString } from './feed/topic'
 import { createFeedManifest } from './modules/feed'
 import { assertBeeUrl, stripLastSlash } from './utils/url'
-import { EthAddress, makeEthAddress, makeHexEthAddress } from './utils/eth'
+import { EthAddress, HexEthAddress, makeEthAddress, makeHexEthAddress } from './utils/eth'
 import { wrapBytesWithHelpers } from './utils/bytes'
 import { assertReference } from './utils/type'
-import { makeJsonFeed } from './feed/json'
+import { setJsonData, getJsonData } from './feed/json'
+import { add } from 'husky'
```
agazso, line 48: This seems to be left in mistakenly.

src/bee.ts
```diff
@@ -41,10 +41,11 @@ import { downloadSingleOwnerChunk, uploadSingleOwnerChunkData } from './chunk/so
 import { makeTopic, makeTopicFromString } from './feed/topic'
 import { createFeedManifest } from './modules/feed'
 import { assertBeeUrl, stripLastSlash } from './utils/url'
-import { EthAddress, makeEthAddress, makeHexEthAddress } from './utils/eth'
+import { EthAddress, HexEthAddress, makeEthAddress, makeHexEthAddress } from './utils/eth'
```
agazso, line 44: Please consider removing the unused definitions.

src/bee.ts
AuHau: Thanks for catching this!

src/bee.ts
```diff
@@ -552,15 +551,31 @@ export class Bee {
...
    */
   getJsonFeed<T extends AnyJson>(topic: string, options?: JsonFeedOptions): Promise<T> {
     const hashedTopic = this.makeFeedTopic(topic)
     const feedType = options?.type ?? DEFAULT_FEED_TYPE
-    const writer = this.makeFeedWriter(feedType, hashedTopic, options?.signer)
-    const jsonFeed = makeJsonFeed<T>(this, writer)
+    let address: EthAddress
+
+    try {
+      address = this.resolveSigner(options?.signer).address
```
agazso, line 563: My expectation when using this function would be that if I provide a value as an option that overrides the value provided in the constructor. Currently that is the case with the `signer` but not with the `address`. Unfortunately it gets even more complicated because it is possible to provide both an optional `signer` and `address` at the same time.

I would suggest to write tests for the different use-cases and use that as a specification and then implement the logic that passes the tests.

AuHau: Yeah, very good point indeed. I was a bit reckless with this implementation, thanks a lot for catching this!

agazso (approved):
Looks good, nice tests!

Issue #265, opened 2021-04-16
Bee class API simplifications


AuHau:
I have discovered few possible improvements:

Currently, there are "two ways" how to download a file, even though they are not related that much internally, but from external API point of view they seems similar. There is `downloadFile` and `downloadFileFromCollection`. I think these could be merged into `downloadFile` where its signature would be extended with optional parameter path to `downloadFile(reference: Reference | string, path?: string)` and if path would be specified than it would use "download file from collection" approach and if not specified than it would use "download file" approach. This would affect both `downloadFile` and `downloadFileReadable`.

~~Another suggestion is actually to drop the `uploadFilesFromDirectory` method and instead provide this functionality through utility function that would build collection and actually create new method `uploadCollection` as I have already proposed in https://github.com/ethersphere/bee-js/issues/172 that you would pass the collection to.~~

Also another thing is that I would like to unify usage of the return type of `Reference` and `ReferenceResponse` choose one or the other and use it consistently. I would lean to use `Reference` I guess. Should verify that the responses does not contain some useful data in headers that should be extracted and passed along like with `File` object.

AuHau:
The consensus is to move forward with the first and third improvements.

AuHau:
Resolved in #280

PR #280, opened 2021-04-22
refactor!: use bzz endpoint for file and dirs

AuHau:
This PR adopts the upcoming changes to bee (see https://github.com/ethersphere/bee/pull/1501). Also tackles partially the #265. It won't be merged atm. but it is ready for review.

Current limitation of this PR:

 - `isSupportedVersion` fails because we are running tests against latest master. 
 - ~~browser `should pin and unpin collection` test is disabled as it now newly uses `Swarm-Collection` header and because of that it runs into bee bug https://github.com/ethersphere/bee/issues/1376~~

Breaking changes:

 - Removing `download*FromCollection`
 - Removed `recursive` flag from `uploadFilesFromDirectory`


src/modules/bzz.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  return headers
+}
+export async function uploadFile(
+  url: string,
+  data: string | Uint8Array | Readable | ArrayBuffer,
+  name?: string,
+  options?: FileUploadOptions,
+): Promise<Reference> {
+  if (!url || url === '') {
```
Cafe137, line 42: I think `''` is falsy

src/modules/bzz.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  return headers
+}
+
+export async function uploadCollection(
+  url: string,
+  collection: Collection<Uint8Array>,
+  options?: CollectionUploadOptions,
+): Promise<Reference> {
+  if (!url || url === '') {
```
Cafe137, line 146: Same here, `''` should be falsy

nugaon (approved):
LGTM!

src/modules/bzz.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  const response = await safeAxios<{ reference: Reference }>({
+    ...options?.axiosOptions,
+    method: 'post',
+    url: url + bzzEndpoint,
+    data: prepareData(data),
+    headers: {
+      ...extractFileUploadHeaders(options),
+    },
+    params,
```
agazso, line 60: It's possible to rewrite this as 
```typescript
params: {
  name,
},
```
Then the `params` declaration above with the `if (name) { ... }` is not necessary.

src/modules/bzz.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  return headers
+}
+export async function uploadFile(
+  url: string,
+  data: string | Uint8Array | Readable | ArrayBuffer,
+  name?: string,
+  options?: FileUploadOptions,
+): Promise<Reference> {
+  if (!url) {
```
agazso, line 42: This check is not strictly necessary because `uploadFile` is not exposed to the public API, therefore it's always called with the inputs already validated.

We also doesn't check the inputs or the url in other functions. If you want stronger guarantees it's possible to create a `BeeURL` type which is already validated and have the type system enforce its validity.

PR #289, opened 2021-04-27
refactor: address type is strict hexstring

AuHau:
Refactor that I need for postage stamp work. Strictly types `Address` type to `HexString`.

### DIFF
--- a/src/bee-debug.ts
+++ b/src/bee-debug.ts
@@ -25,7 +25,7 @@ import type {
   NodeAddresses,
 } from './types'
 import { assertBeeUrl, stripLastSlash } from './utils/url'
-import { assertInteger } from './utils/type'
+import { assertAddress, assertInteger } from './utils/type'
 
 /**
  * The BeeDebug class provides a way of interacting with the Bee debug APIs based on the provided url
@@ -87,7 +87,9 @@ export class BeeDebug {
    *
    * @param address Swarm address of peer
    */
-  getPeerBalance(address: Address): Promise<PeerBalance> {
+  getPeerBalance(address: Address | string): Promise<PeerBalance> {
+    assertAddress(address)
+
     return balance.getPeerBalance(this.url, address)
   }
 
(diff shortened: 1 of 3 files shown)


src/bee-debug.ts
```diff
@@ -25,7 +25,7 @@ import type {
   NodeAddresses,
 } from './types'
 import { assertBeeUrl, stripLastSlash } from './utils/url'
-import { assertInteger } from './utils/type'
+import { assertAddress, assertInteger } from './utils/type'
```
agazso, line 28: In the codebase we have assertion functions defined as `assertType` but also `assertIsType`. It seems that the latter is the more widespread naming when using with type names, at least in the [Typescript documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) and [here](https://2ality.com/2020/06/type-guards-assertion-functions-typescript.html#asserting-the-type-of-an-argument%3A-asserts-«arg»-is-«type»).

It would be good to be consistent in this regard.

AuHau: Hmmm, honestly I am not aware of any `assertIsType` function, can you give an example, please?

agazso: `assertIsFeedType`, `assertIsEthAddress` and `assertIsSwarmNetworkId`

AuHau: Oh interesting, I did not find it. You are right and agree that we should unify this, but I would more incline to use the `assertType` style.

nugaon (approved):

nugaon:
next time please target that branch your PR is based on.

AuHau:
@nugaon sorry, that was because of some rebases that I did not propagate all the way here.

src/bee-debug.ts
AuHau: https://github.com/ethersphere/bee-js/issues/296

PR #290, opened 2021-04-27
feat: postage stamp support

AuHau:
Initial work on Postage stamp support. It is still very much WIP without any tests.


src/bee.ts
```diff
@@ -515,10 +535,42 @@ export class Bee {
...
   }
 
+  /**
+   * Creates new postage batch from the funds that the node has available.
+   *
+   * @param amount Amount added to the balance of the batch
+   * @param depth Logarithm of the number of chunks that can be stamped with the batch
+   * @param label An optional label for the batch
+   */
+  createStampBatch(amount: bigint, depth: number, label?: string): Promise<Address> {
```
agazso, line 549: I find the naming of this a bit strange, considering that `Stamp` and `Batch` is used interchangeably in Swarm. This is quite confusing already so it would be better to choose one of those and use it consistently in the API.

Therefore I think calling this `createPostageStamp` or `createPostageBatch` would be better.

src/bee.ts
```diff
@@ -541,6 +593,22 @@ export class Bee {
...
+   */
+  getDefaultPostageBatchId(): Address | undefined {
+    return this.postageBatchId
+  }
+
+  /**
+   * Setter for default postage batchId
+   * @param batchId
+   */
+  setDefaultPostageBatchId(batchId: Address | string): void {
```
agazso, line 607: Is the setter really necessary? I would prefer to keep the `Bee` class readonly.

src/chunk/soc.ts
```diff
@@ -126,43 +127,60 @@ export async function makeSingleOwnerChunk(
...
+ * @param signer          The singer interface for signing the chunk
+ * @param postageBatchId
+ * @param identifier      The identifier of the chunk
+ * @param data            The chunk data
  * @param options
  */
 export async function uploadSingleOwnerChunkData(
   url: string,
   signer: Signer,
+  postageBatchId: Address | undefined,
```
agazso, line 162: I find having two optional arguments confusing, maybe can we rid of one of them?

src/utils/headers.ts
```diff
@@ -54,8 +55,14 @@ export function readFileHeaders(headers: Dictionary<string>): FileHeaders {
   }
 }
 
-export function extractUploadHeaders(options?: UploadOptions): UploadHeaders {
-  const headers: UploadHeaders = {}
+export function extractUploadHeaders(postageBatchId: Address, options?: UploadOptions): UploadHeaders {
```
agazso, line 58: It seems an error that `UploadOptions` has a `postageBatchId` but it is not used in this function. Or if it's intentional it is very confusing and at least should be documented.

src/utils/type.ts
```diff
@@ -15,6 +15,12 @@ export function assertInteger(value: unknown): asserts value is number | bigint
   if (!isInteger(value)) throw new TypeError('value is not integer')
 }
 
+export function assertPositiveInteger(value: unknown): asserts value is number | bigint {
```
agazso, line 18: This function returns positive integers **and** zero. Maybe it would be better to call it `assertNonNegativeInteger`?

agazso (commented):
I find the handling of the optional `postageBatchId` arguments mixed with the also optional `UploadOptions` `postageBatchId` very confusing.

I think at the module level we can just treat it as a required argument and assert in the input validation that it was provided.

src/bee.ts
AuHau: I knew you would object to this and I agree with your thinking, but I think it is necessary. 

First of all, it is necessary to have access to `Bee` class to actually create/retrieve the stamps, so for example for the first time running application you would have to create a temporary `Bee` instance to only create the postage stamp and only then you could create the "real" `Bee` instance with the default stamp set. I would argue this is too cumbersome.  

Also, I think the postage batches are quite dynamic. You might have to switch between them, create a new one, and so on and whenever you would like to change the default batch stamp, then you would have to recreate the `Bee` class which again I think is too cumbersome.

src/bee.ts
AuHau: Yeah, I myself am quite confused by the mixing terminology in Bee. I will improve this 👍

src/bee.ts
vojtechsimetka:
```suggestion
  set defaultPostageBatchId(batchId: Address | string): void {
```
Can't we make this a real setter?

I also prefer not to have this. If user wants to have such abstraction, they can do this on their own extending the Bee class

agazso (approved):
I think this is good for now and later, once we have a better understanding of how the postage stamps are actually used in applications, then we may think of adding convenience functions around it.

The README should be changed after this is merged, but it can be part of a separate PR before the release.

src/types/index.ts
```diff
@@ -202,7 +217,20 @@
    * @param data        The chunk payload data
    * @param options     Upload options
    */
-  upload: (identifier: Identifier, data: Uint8Array, options?: UploadOptions) => Promise<ReferenceResponse>
+  upload: (
+    postageBatchId: string | Address,
+    identifier: Identifier,
+    data: Uint8Array,
+    options?: UploadOptions,
```
nugaon, line 224: same as I mentioned previously, I don't think we should interpret the same as the `Address`es

src/modules/stamps.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+interface CreateStampResponse {
+  batchID: Address
+}
+
+export async function getAllPostageBatches(url: string): Promise<StampBatch[]> {
+  const response = await safeAxios<GetAllStampsResponse>({
+    method: 'get',
+    url: url + STAMPS_ENDPOINT,
```
nugaon, line 17: this is no problem at all, but can you just keep the same format as in other cases? (template literals)

src/utils/headers.ts
nugaon: yes, I second the opinion of @agazso, the `postageBatchId` seems always required at headers additions, I wouldn't keep it in `UploadOptions`

src/types/index.ts
```diff
@@ -54,6 +60,7 @@
   'swarm-pin'?: string
   'swarm-encrypt'?: string
   'swarm-tag'?: string
+  'swarm-postage-batch-id'?: string
```
nugaon, line 63: So as we mentioned earlier it seems to be unnecessary, because `extractUploadHeaders` already has as a required parameter.

src/types/index.ts
```diff
@@ -84,7 +95,7 @@
~export interface Data extends Uint8Array {
~  text(): string
~  hex(): HexString
~  json(): Record<string, any>
~}
```
nugaon: not updated jsdoc

src/bee.ts
```diff
@@ -74,13 +82,16 @@ export class Bee {
...
    *
+   * @param postageBatchId Postage BatchId to be used to upload the data with
    * @param data    Data to be uploaded
    * @param options Additional options like tag, encryption, pinning, content-type
    *
    * @returns reference is a content hash of the data
    */
-  uploadData(data: string | Uint8Array, options?: UploadOptions): Promise<Reference> {
-    return bytes.upload(this.url, data, options)
+  uploadData(postageBatchId: string | Address, data: string | Uint8Array, options?: UploadOptions): Promise<Reference> {
```
nugaon, line 91: instead of `string | Address` I would use full flavourtypes, or it makes harder to understand for users in general?

src/bee.ts
```diff
@@ -230,148 +225,72 @@ export class Bee {
...
   /**
    * Send to recipient or target with Postal Service for Swarm
    *
+   * @param postageBatchId Postage BatchId to be used to upload the data with
    * @param topic Topic name
    * @param target Target message address prefix
    * @param data Message to be sent
    * @param recipient Recipient public key
-   *
+   * @param postageBatchId Postage BatchId that will be assigned to sent message
```
nugaon, line 282: remove this line pls

test/utils.ts
```diff
@@ -115,6 +155,33 @@ export function beePeerUrl(): string {
...
+    try {
+      batchId[url] = await createPostageBatch(url, BigInt('1000'), 25)
+    } catch (e) {
+      await sleep(500)
+
+      const batches = await getAllPostageBatches(url)
+
+      if (!batches.length) {
+        batchId[url] = await createPostageBatch(url, BigInt('1000'), 25)
+      } else {
```
nugaon, line 176: I'm planning to create postage batch on blockchain image build in `bee-factory`, so later this could be replaced with a system environment variable which contains this "monster" postage batch ID : )

test/integration/modules/pinning.spec.ts
```diff
@@ -11,39 +10,49 @@ import {
...
     it('should unpin an existing file', async () => {
-      const hash = await file.upload(BEE_URL, randomData)
-      const response = await pinning.unpinFile(BEE_URL, hash)
+      const hash = await bzz.uploadFile(BEE_URL, randomData, await getPostageBatch())
+      const response = await pinning.unpin(BEE_URL, hash)
 
       expect(response).toEqual(okResponse)
     })
 
+    // TODO: https://github.com/ethersphere/bee-js/issues/304
```
nugaon, line 45: it has been resolved as I checked your comment

test/integration/modules/bzz.spec.ts
```diff
@@ -0,0 +1,284 @@
+import * as bzz from '../../../src/modules/bzz'
+import { Collection, ENCRYPTED_REFERENCE_HEX_LENGTH } from '../../../src/types'
+import {
+  beeUrl,
+  BIG_FILE_TIMEOUT,
+  createReadable,
+  ERR_TIMEOUT,
```
nugaon, line 7: not used import

nugaon (commented):
Instead of `assertAddress` and `Address` I would use different type for batchIds as they are not ethereum addresses or even chunk addresses, as you can see here: https://github.com/ethersphere/storage-incentives/blob/master/src/PostageStamp.sol#L74

src/types/index.ts
AuHau: This is actually correct, because the `extractUploadHeaders` returns the postage batch as  header. This is not the `UploadOptions` that yes it was removed from.

AuHau:
Thanks @nugaon for the comments. I have created an issue about the `Address` type change and I will leave it as it is now for the upcoming release and will revisit it once it is out.

PR #293, opened 2021-04-28
refactor: new pinning api

### LINKED ISSUE #292: Pinning breaking changes
Bee simplified the pinning API in https://github.com/ethersphere/bee/pull/1566 we should adapt those changes.

AuHau:
Closes #292 

Breaking changes:

 - Following methods are removed `bee.pinFile()`, `bee.unpinFile()`, `bee.pinCollection()` `bee.unpinCollection()`, `bee.pinData()`, `bee.unpinData()`, `bee.pinChunk()`, `bee.unpinChunk()`, `bee. getChunkPinningStatus()`

### DIFF
--- a/src/modules/pinning.ts
+++ b/src/modules/pinning.ts
@@ -4,7 +4,7 @@ import { safeAxios } from '../utils/safeAxios'
 const PINNING_ENDPOINT = '/pins'
 
 export interface GetAllPinResponse {
-  chunks: Pin[]
+  references: Pin[] | null
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/modules/pinning.ts
```diff
@@ -1,185 +1,72 @@
...
     method: 'get',
     responseType: 'json',
-    url: `${url}${Endpoint.CHUNKS}`,
-    params: options,
+    url: `${url}${PINNING_ENDPOINT}`,
   })
 
-  return response.data
+  return response.data.chunks
 }
```
Cafe137, line 72: I believe when no pins are available, the Bee endpoint returns `null`. Can you please verify this behaviour, and in that case, maybe convert it to an empty array so that the signature holds up?

agazso (approved):

PR #297, opened 2021-05-05
ci: bee-factory

nugaon:
This PR changes the environment of the JS tests to the [Bee Factory](https://github.com/ethersphere/bee-factory) docker environment.

Fixes: #298

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -40,6 +40,11 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
+      - name: Use Python 2.7 for setups
+        uses: actions/setup-python@v2
+        with:
+        python-version: '2.7'
+
       - name: Auth to Github Package Docker Registry
         run: |
           echo "${{ secrets.GITHUB_TOKEN }}" | docker login https://docker.pkg.github.com -u ${GITHUB_ACTOR} --password-stdin
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -95,6 +95,15 @@ jobs:
         with:
           fetch-depth: 1
 
+      - name: Use Python 2.7 for setups
+        uses: actions/setup-python@v2
+        with:
+          python-version: '2.7'
+
+      - name: Auth to Github Package Docker Registry
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | docker login https://docker.pkg.github.com -u ${GITHUB_ACTOR} --password-stdin
+
       # Setup Bee environment
       - name: Start Bee Factory environment
         run: |
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -40,6 +40,10 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
+      - name: Auth to Github Package Docker Registry
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | docker login https://docker.pkg.github.com -u ${GITHUB_ACTOR} --password-stdin
+
       # Setup Bee environment
       - name: Start Bee Factory environment
         run: |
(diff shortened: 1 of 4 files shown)


nugaon:
I had to change the timeouts on the feed updates from 15 sec to 21, because it always failed. 
if the feed lookup will be more optimized, we can decrease this number

nugaon:
wrong test on integration/modules/debug/connectivity.spec.ts
  ● modules/debug/connectivity › pingPeer

    Expected pattern: /^\d+(\.\d+)[mnp]?s$/
    Received string:  "924.671µs"

the environment is faster than it is expected. : )

Cafe137 (approved):
LGTM!

.github/workflows/tests.yaml
```diff
@@ -37,30 +40,21 @@ jobs:
...
-      - name: Set testing cluster
+          echo "${{ secrets.GITHUB_TOKEN }}" | docker login https://docker.pkg.github.com -u ${GITHUB_ACTOR} --password-stdin
+
+      # Setup Bee environment
+      - name: Start Bee Factory environment
         run: |
-          echo -e "127.0.0.10\tregistry.localhost" | sudo tee -a /etc/hosts
-          for ((i=0; i<REPLICA; i++)); do echo -e "127.0.1.$((i+1))\tbee-${i}.localhost bee-${i}-debug.localhost"; done | sudo tee -a /etc/hosts
-          ./beeinfra.sh install -r "${REPLICA}" --geth
+          git clone https://github.com/ethersphere/bee-factory.git
```
agazso, line 55: I think we could use `git clone --depth=1` for quicker clone operation in this case.

For more details see for example this article:
https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/

.github/workflows/tests.yaml
```diff
@@ -101,30 +95,21 @@ jobs:
...
```
agazso, line 110: See my other comment regarding shallow cloning.

.github/workflows/tests.yaml
```diff
@@ -37,30 +40,21 @@ jobs:
...
-          export TAG=$(curl -s https://api.github.com/repos/ethersphere/beekeeper/releases/latest | jq -r .tag_name)
-          curl -Ls https://github.com/ethersphere/beekeeper/releases/download/${TAG}/beekeeper-linux-amd64 -o beekeeper
-          chmod +x beekeeper
-      - name: Prepare testing cluster
-        run: ./beeinfra.sh prepare --geth
-      - name: Set kube config
+      - name: Use Python 2.7 for setups
+        uses: actions/setup-python@v2
+        with:
+          python-version: '2.7'
```
agazso, line 46: Why do we need Python 2.7? Python version 2 is no longer supported since 2020. In most cases it's trivial to change the code to use Python 3.

agazso (approved):
Looks good, but I wouldn't use Python 2 anymore

.github/workflows/tests.yaml
nugaon: because many computer support only python 2.7 by default still or that is the default python version.

as PR #12 got approves from devs who use OSX, I think this is the case on Apple computers as well.

it is not a python project, I don't think it would be a big problem, only reason we need python 2 because of [the print syntax](https://github.com/ethersphere/bee-factory/blob/5b3eba940008390f8d193121eee4c6c6b01f4412/scripts/bee.sh#L60) is different between python 2 and 3.

agazso: `print x` only works in Python 2
`print(x)` works both in Python 2 and 3

I suggest to change it and then it will work if we change it to Python 3 as well.

AuHau (approved):
LGTM, I like how it simplified the setup! But I agree with Attila on using Python 3.

AuHau:
Btw. I don't think this is a breaking PR (from the bee-js API perspective) so I would remove `!` from the final commit ;-)

PR #312, opened 2021-05-19
feat: gas price for postage batch creation

AuHau:
This add functionality to specify Gas Price for the creation of postage batch as introduced in https://github.com/ethersphere/bee/pull/1769


Cafe137 (approved):
LGTM

agazso:
The test is failing because the check for the short git hash expects exactly 8 hex characters.

However based on the answer of [this](https://stackoverflow.com/questions/5694389/get-the-short-git-version-hash) it seems that on some systems the minimum length is 7 characters and for git the minimum is actually 4 characters.

Please note also that based on the [documentation of git](https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt---shortlength) this is only the _minimum_ length, but since git returns a unique short hash it may be longer than that.

Based on this I would suggest to change the `{8}` to `{4,}` (4 or more) in the regex.

AuHau:
Thanks Attila! Good suggestion. Yeah this PR is based on version of `bee-0.6.0` before it had the regex fix.

PR #314, opened 2021-05-20
feat: pass error message from bee to thrown error

### LINKED ISSUE #303: Improve error reporting
Currently, when something goes wrong on Bee node we return the `BeeResponseError` error that has a status code and status text. I would argue that status text does not give the user any valuable information. 

Bee actually returns quite often object with `message` and `code` property, where `message` usually contains some more pointers on what went wrong (even though it is not much verbose). I would hence propose to replace the status text with the `message` property (if it is included in the response).

AuHau:
Closes #303

### DIFF
--- a/src/utils/safeAxios.ts
+++ b/src/utils/safeAxios.ts
@@ -103,7 +103,11 @@ export async function safeAxios<T>(config: AxiosRequestConfig & { forceBigInt?: 
     return response
   } catch (e) {
     if (e.response) {
-      throw new BeeResponseError(e.response.status, e.response.statusText)
+      if (e.response.data?.message) {
+        throw new BeeResponseError(e.response.status, `${e.response.statusText}: ${e.response.data.message}`)
+      } else {
+        throw new BeeResponseError(e.response.status, e.response.statusText)
+      }
     } else if (e.request) {
       throw new BeeRequestError(e.message)
     } else {


src/utils/safeAxios.ts
```diff
@@ -103,7 +103,11 @@ export async function safeAxios<T>(config: AxiosRequestConfig & { forceBigInt?:
     return response
   } catch (e) {
     if (e.response) {
-      throw new BeeResponseError(e.response.status, e.response.statusText)
+      if (e.response.data?.message) {
+        throw new BeeResponseError(e.response.status, `${e.response.statusText}: ${e.response.data.message}`)
+      } else {
+        throw new BeeResponseError(e.response.status, e.response.statusText)
```
nugaon, line 109: I think the error message formatting should be done in `BeeResponseError` based on the passed parameters.

AuHau: Hmm, honestly I don't see a reason for that.

nugaon: If you want to use elsewhere this error class, then you don't have to remember the formatting style that you used before and in need you can change it in one place.

agazso: I would also not merge the `statusText` (which is a HTTP response message) with the actual response from the Bee node in a string.

How about adding a `message` property to the `BeeResponseError` and store it there?

AuHau: This parameter we are talking about is the message. It is then passed into `Error` as the message which is then shown when the Error is converted to string (eq. in logs etc). Having the `statusText` and `response.data.message` separate does not really make sense. If you need to do some assertions than you should use `status` (error code) which is passed separately and not on the `statusText`. This is mainly for people's eyes.

Errors in my view are simple vessels that only carry information and do not really perform any formatting and other logic.

AuHau: @ethersphere/js what do you think? From my side with a bit of distance from this, I would still keep it as is. 

A possible different solution could be not to include the `statusText` and instead then internally in the Error class prepend the `status` code which would lead to something like `400: bad postage stamp` or `400: Bad Request` if the `message` is not specified.

agazso: I don't have a too strong opinion about either way, just please keep in mind that in the swarm-cli currently we depend on the value of the error message to detect certain errors and changing that would break existing behavior there.

https://github.com/ethersphere/swarm-cli/issues/117

nugaon: okay, I don't want to argue on this, we can keep it as it is now.

Cafe137 (approved):
LGTM!

nugaon (approved):

PR #318, opened 2021-05-20
feat: limit postage depth

AuHau:
When implementing the Postage batch support I forgot that depth is limited to at least 16. This adds a check for this.

### DIFF
--- a/src/bee.ts
+++ b/src/bee.ts
@@ -50,7 +50,6 @@ import { assertAddress, assertNonNegativeInteger, assertReference } from './util
 import { setJsonData, getJsonData } from './feed/json'
 import { makeCollectionFromFS, makeCollectionFromFileList } from './utils/collection'
 import { PostageBatchOptions } from './types'
-import { MINIMUM_DEPTH } from './modules/stamps'
 
 /**
  * The Bee class provides a way of interacting with the Bee APIs based on the provided url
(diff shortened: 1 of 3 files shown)


src/bee.ts
```diff
@@ -576,6 +577,10 @@ export class Bee {
     assertNonNegativeInteger(amount)
     assertNonNegativeInteger(depth)
 
+    if (depth < stamps.MINIMUM_DEPTH) {
```
agazso, line 580: `MINIMUM_DEPTH` is imported from `./modules/stamps`, so you either don't need to specify the module here or you don't need the import.

AuHau: I don't understand what you meant. I just wanted to place the constant somewhere appropriately so chose that module, but the input validation should happen here.

agazso: There is a separate import line importing only `MINIMUM_DEPTH` (line 53), but there is an unqualified import for the `stamps` module already.

I suggest you to remove line 53.

src/bee.ts
```diff
@@ -576,6 +577,10 @@ export class Bee {
...
+      throw new BeeArgumentError('Depth has to be greater or equal then 16', depth)
```
agazso, line 581: Typo:
```suggestion
      throw new BeeArgumentError('Depth has to be greater or equal than 16', depth)
```

Cafe137:
Maximum depth is 255. 256 fails with HTTP 400.

src/bee.ts
Cafe137: > Typo:

`greater than or equal to` would be correct, no?

Cafe137: > Typo:

```
'Depth has to be greater than or equal to ' + MINIMUM_DEPTH
```

or

```
'Depth has to be at least ' + MINIMUM_DEPTH
```

src/bee.ts
AuHau: Damn, my idea is too happy to import stuff without my awareness 😅 Thanks for the catch!

agazso (approved):

Cafe137 (approved):

PR #319, opened 2021-05-20
fix: if there are no postage stamps the getAllPostageBatch should return []

vojtechsimetka:
No idea how to test this since the tests are in random order and I can not guarantee there is no postage stamp, right?

### DIFF
--- a/src/bee.ts
+++ b/src/bee.ts
@@ -605,7 +605,7 @@ export class Bee {
   /**
    * Return all postage batches that has the node available.
    */
-  getAllPostageBatch(): Promise<PostageBatch[]> {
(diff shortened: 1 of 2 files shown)


AuHau:
Thanks! Good catch! I fixed this for the pinning endpoint but forgot that this will be the case here as well.

I think that actually, the approach here is to check if null is returned and then return empty array instead.

https://github.com/ethersphere/bee-js/blob/bee-0.6.0/src/modules/pinning.ts#L73

AuHau:
Haha, yeah, you were faster.

AuHau (approved):
LGTM

vojtechsimetka:
This should actually be even simpler because `null || []` results in `[]`

PR #323, opened 2021-05-21
feat: reupload

### LINKED ISSUE #313: Reupload of data
There is a new upcoming feature of Bee that allows reuploading data that is locally pinned. We should expose this.

https://github.com/ethersphere/bee/commit/8f8c05fa025db892a13a27dbee57df649955e927

AuHau:
Closes #313

### DIFF
--- a/src/modules/bzz.ts
+++ b/src/modules/bzz.ts
@@ -182,3 +182,19 @@ export async function uploadCollection(
 
   return response.data.reference
 }
+
+/**
+ * Reupload locally pinned data
+ * @param url
+ * @param reference
+ * @param options
+ * @throws BeeResponseError if not locally pinned or invalid data
+ */
+export async function reupload(url: string, reference: Reference, options?: AxiosRequestConfig): Promise<void> {
+  await safeAxios({
+    ...options,
+    method: 'patch',
+    url: `${url}${bzzEndpoint}/${reference}/`,
+    responseType: 'json',
+  })
+}
--- a/src/bee.ts
+++ b/src/bee.ts
@@ -272,6 +272,22 @@ export class Bee {
     return pinning.getPin(this.url, reference)
   }
 
+  async reuploadPinnedData(reference: Reference | string): Promise<void> {
+    assertReference(reference)
+
+    try {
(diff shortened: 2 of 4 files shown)


src/modules/bzz.ts
```diff
@@ -182,3 +182,19 @@ export async function uploadCollection(
...
 }
+
+/**
+ * Reupload locally pinned data
+ * @param url
+ * @param reference
+ * @param options
+ * @throws BeeResponseError if not locally pinned or invalid data
+ */
+export async function reupload(url: string, reference: Reference, options?: AxiosRequestConfig): Promise<void> {
```
agazso, line 193: Why there is an `options` argument? It doesn't seem to be used anywhere.

AuHau: Well, this was taken from the way how other modules are also implemented which allows overriding the axios configuration. I forgot to expose it in Bee class. But I am not sure if we should actually allow overriding configuration of axios like this? What is the reasoning behind it?

nugaon (commented):
The code looks good, regarding to the axiosOptions on module level we can provide this additionally but I also don't see real any use-case now for that now.
for [the Bee issue](https://github.com/ethersphere/bee/issues/1803) I'd rather propose using 422 http error, as the syntax and the request are in great format and understandable by the "server", but the business-logic does not let reupload an unpinned content.

AuHau:
> The code looks good, regarding to the axiosOptions on module level we can provide this additionally but I also don't see real any use-case now for that now.

@nugaon you mean for this reupload endpoint or generally in Bee JS?

> for the Bee issue I'd rather propose using 422 http error, as the syntax and the request are in great format and understandable by the "server", but the business-logic does not let reupload an unpinned content.

Hmm not sure how it is related to this? I would suggest to comment in the linked issue as this is up to the Bee team to decide

nugaon:
> @nugaon you mean for this reupload endpoint or generally in Bee JS?

I meant I don't see the use-case for this reupload endpoint regarding to the axios options. anyway it can be unified as in our other modules we do the same, it doesn't pollute the code.

nugaon:
> Hmm not sure how it is related to this? I would suggest to comment in the linked issue as this is up to the Bee team to decide

it was commented in your code in this PR.

PR #324, opened 2021-05-24
feat: chain and reserve state endpoints

### LINKED ISSUE #317: Bee Debug: new endpoints ReserveState and ChainState
There are new endpoints on the Bee Debug endpoint which should be exposed.

https://github.com/ethersphere/bee/commit/88c0a1955ecd1e87ff84b6a456cf10a1bbbbfc2d#diff-a679fc0102a02684f0e2be820349e2de2172ce98b42f0e4f4c08825f53d9010fR252

AuHau:
Closes #317

### DIFF
--- a/src/bee-debug.ts
+++ b/src/bee-debug.ts
@@ -3,6 +3,7 @@ import * as balance from './modules/debug/balance'
 import * as chequebook from './modules/debug/chequebook'
 import * as settlements from './modules/debug/settlements'
 import * as status from './modules/debug/status'
+import * as states from './modules/debug/states'
 import type {
   Address,
   Peer,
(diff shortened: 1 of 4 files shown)


nugaon (approved):
LGTM!

AuHau:
@nugaon what do you think? Currently, the `block` in the `chainState` endpoint is also `BigInt` even though it does not have to be, but because I use the `forceBigInt` flag axios parse all numbers into `bigint`s. Do you think it is a problem?

nugaon:
yeah, I noticed, I'm not bothered by that : ) though if this does'nt make you big difficulty you can keep it as number.

PR #325, opened 2021-05-25
feat!: support gas price and limit for cashout

### LINKED ISSUE #310: Cashout cheque gas-price/limit
The cashout endpoint supports specifying gas price and gas limit, we currently do not expose these options in `BeeDebug`. FYI. these should be `BigInt`s

AuHau:
Closes #310 

Honestly, I would like to refactor the `cashoutLastCheque` to return the `transactionHash` directly and not wrapped in an object as is now, but it would be breaking change. What do you think?

### DIFF
--- a/src/bee-debug.ts
+++ b/src/bee-debug.ts
@@ -25,7 +25,9 @@ import type {
   NodeAddresses,
 } from './types'
 import { assertBeeUrl, stripLastSlash } from './utils/url'
-import { assertAddress, assertInteger } from './utils/type'
+import { assertAddress, assertInteger, assertNonNegativeInteger } from './utils/type'
+import { CashoutOptions } from './types'
+import { add } from 'husky'
 
 /**
  * The BeeDebug class provides a way of interacting with the Bee debug APIs based on the provided url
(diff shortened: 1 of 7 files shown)


src/bee-debug.ts
```diff
@@ -25,7 +25,9 @@ import type {
   NodeAddresses,
 } from './types'
 import { assertBeeUrl, stripLastSlash } from './utils/url'
-import { assertAddress, assertInteger } from './utils/type'
+import { assertAddress, assertInteger, assertNonNegativeInteger } from './utils/type'
+import { CashoutOptions } from './types'
+import { add } from 'husky'
```
nugaon, line 30: not used import
```suggestion
```

nugaon (approved):
looks good and I second your suggestion about the unwrapped return value.

nugaon:
Also, there is unused import in `bee-class.spec.ts`  (`BeeError`), and it has type issue for the given `gasPrice` in line 156

AuHau:
Thanks for the feedback, that is my trigger happy IDE 😅

PR #326, opened 2021-05-25
fix!: correctly return reject promise for promise returning fnc

### LINKED ISSUE #311: Mixing throwing errors and Promise rejections
In `Bee` and `BeeDebug` we have a lot of functions that return Promise but themselves are not `async`. If there are some assertions in these functions that usually throw errors, these errors are then not converted to rejected Promise but are thrown as actually Errors. 

Even though it is valid action, I would argue that it is not expected by the developers as they could expect rejected Promise. (I would)

If the devs use the `async/await` paradigm, then it is not a problem as they would use `try/catch` for catching the Errors which should catch both rejected promises as well as normal errors. But if the devs are using Promise API, then this can pass their error handling as `ourFunction().catch(...)` won't be invoked.

What are your thoughts?
@ethersphere/js

AuHau:
Closes #311.

This PR is breaking as the previous behavior was throwing errors and now will be returning rejected Promise instead.

As part of the PR I have updated ESLint dependencies, bit modified the `ban-ts-command` rule, and did a general sweep of the codebase to remove ESlint warnings.


.eslintrc.js
```diff
@@ -41,7 +42,6 @@ module.exports = {
     'require-yield': 'error',
     'max-nested-callbacks': ['error', 4],
     'max-depth': ['error', 4],
-    'require-await': 'error',
```
nugaon, line 44: Why did you remove this rule? With this everything works fine as you stated in this PR the goal is to return back from `async` functions always promise.

nugaon, line 44: Why did you delete this rule? with this everything works fine as you stated in this PR the goal is always return promise in case of `async` functions, and [this rule assures that](https://eslint.org/docs/rules/require-await) -> `The return value is always a Promise.`

AuHau: I have not removed it but turned it off below. And no, everything does not work fine with this rule. This rule "Disallow async functions which have no await expression" which then leads to not using `async` and hence throwing exceptions instead of returning the Promise and to the whole reason that I am trying to fix here.

AuHau: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/promise-function-async.md

This is the new rule that enforces the new behavior.

AuHau: Hmm I originally thought, that this new rule required the turning off the `require-await` but I mixed that up with this rule: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/require-await.md

I am gonna test the Typescript's version of `require-await` to see if it would work correctly, but I don't think the original `require-await` works.

nugaon (approved):

PR #328, opened 2021-05-26
refactor: custom type for postage batch

### LINKED ISSUE #309: Find better type for `postageBatchId`
As @nugaon pointed out in #290 

> Instead of assertAddress and Address I would use different type for batchIds as they are not ethereum addresses or even chunk addresses, as you can see here: ethersphere/storage-incentives@master/src/PostageStamp.sol#L74

AuHau:
Closes #309.

I am not sure if the `BatchId` should be branded or not. What do you think? 🤔

### DIFF
--- a/src/bee.ts
+++ b/src/bee.ts
@@ -30,7 +30,6 @@ import type {
   JsonFeedOptions,
   AnyJson,
   Pin,
-  Address,
   PostageBatch,
   BatchId,
 } from './types'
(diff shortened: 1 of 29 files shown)


src/bee.ts
```diff
@@ -32,6 +32,7 @@ import type {
   Pin,
   Address,
```
agazso, line 33: Please remove unused imports (like this)

agazso (approved):
Looks good but please remove unused imports (see warnings from Github Actions or run checks locally)

vojtechsimetka (approved):

nugaon (commented):
> I am not sure if the `BatchId` should be branded or not. What do you think?

it can be even branded, because you have to cast it anyway to satisfy `HexString`'s definition for 
`string & {
    readonly length: Length
  }`

PR #338, opened 2021-06-02
chore: npm ci script

### LINKED ISSUE #18: Make the CI job easy to run locally
Ideally as easy as `npm run ci`

Currently it is running with `npm ci` but that removes `node_modules` first and then install all the packages.

AuHau:
Closes #18

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "compile:node": "webpack --progress --env target=node",
     "compile:types": "tsc --emitDeclarationOnly --declaration",
     "compile:browser": "webpack --progress --env target=web",
+    "ci": "npm run lint && npm run check:types && npm run test",
     "docs": "rimraf docs && typedoc",
     "test": "npm run compile:browser && jest --verbose --selectProjects=node:unit node:integration dom:integration --config=jest.config.ts",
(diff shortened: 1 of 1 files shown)


agazso (approved):

nugaon (commented):
I suggest that we should rename this `ci` keyword because it is not related to `npm ci` and also it is not used by the CI.

AuHau:
@nugaon I guess I agree with you, but I don't have any ideas on what name use. Do you?

package.json
```diff
@@ -33,6 +33,7 @@
     "compile:node": "webpack --progress --env target=node",
     "compile:types": "tsc --emitDeclarationOnly --declaration",
     "compile:browser": "webpack --progress --env target=web",
+    "ci": "npm run lint && npm run check:types && npm run test",
```
nugaon, line 36: I wonder if you only check the types, why you don't check the lint as well instead of make the corrections?

```suggestion
    "ci": "npm run lint:check && npm run check:types && npm run test",
```

nugaon (commented):
> @nugaon I guess I agree with you, but I don't have any ideas on what name use. Do you?

maybe `mock-ci`?

PR #341, opened 2021-06-03
refactor!: no single-property object returned

### LINKED ISSUE #163: Refactor Writers to return `Reference` instead of `ReferenceResponse`
Currently, the `SOCWriter` and `FeedWriter` function `upload` returns `ReferenceResponse` (eq. object `{ reference: string }`), while most of the "write" APIs on `Bee` class actually returns `Reference` (eq. the hash string) directly. For sake of consistency, we might consider refactoring the writers to follow this pattern.

AuHau:
Closes #163

### DIFF
--- a/test/integration/feed/index.spec.ts
+++ b/test/integration/feed/index.spec.ts
@@ -1,5 +1,5 @@
 import { fetchFeedUpdate } from '../../../src/modules/feed'
-import { HexString, hexToBytes, makeHexString } from '../../../src/utils/hex'
+import { hexToBytes, makeHexString } from '../../../src/utils/hex'
 import { beeUrl, ERR_TIMEOUT, getPostageBatch, testIdentity } from '../../utils'
 import { ChunkReference, downloadFeedUpdate, findNextIndex, Index, uploadFeedUpdate } from '../../../src/feed'
 import { Bytes, assertBytes } from '../../../src/utils/bytes'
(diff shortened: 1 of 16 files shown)


test/unit/feed/json.spec.ts
```diff
@@ -12,7 +12,7 @@ interface CircularReference {
 
 describe('JsonFeed', () => {
   const DATA_REFERENCE = testChunkHash as Reference
-  const FEED_REFERENCE_HASH = 'ca6357a08e317d15ec560fef34e4c45f8f19f01c372aa70f1da72bfa7f1a1111'
+  const FEED_REFERENCE_HASH = 'ca6357a08e317d15ec560fef34e4c45f8f19f01c372aa70f1da72bfa7f1a1111' as Reference
   const FEED_REFERENCE = {
```
vojtechsimetka, line 16: Does this still need to exist? Shouldn't it always just be `FEED_REFERENCE_HASH`?

AuHau: Yes, because the `FeedWriter.download()` returns an object like this as it contains other properties.

test/integration/feed/index.spec.ts
```diff
@@ -15,9 +15,9 @@ function makeChunk(index: number) {
 
 async function uploadChunk(url: string, index: number): Promise<ChunkReference> {
   const chunk = makeChunk(index)
-  const referenceResponse = await chunkAPI.upload(url, chunk.data, getPostageBatch())
+  const reference = await chunkAPI.upload(url, chunk.data, getPostageBatch())
 
-  return hexToBytes(referenceResponse.reference as HexString) as ChunkReference
+  return hexToBytes(reference as HexString) as ChunkReference
```
nugaon, line 20: the type assertion is not not necessary here.
```suggestion
  return hexToBytes(reference) as ChunkReference
```

nugaon (approved):

PR #342, opened 2021-06-03
refactor!: no generic BeeResponse returned from Bee class

AuHau:


test/integration/bee-class.spec.ts
```diff
@@ -127,8 +125,7 @@ describe('Bee class', () => {
       const content = new Uint8Array([1, 2, 3])
       const hash = await bee.uploadFile(getPostageBatch(), content)
 
-      const pinResponse = await bee.pin(hash)
-      expect(pinResponse).toBeOneOf([createdResponse, okResponse])
+      await bee.pin(hash) // Nothing is returned, will throw error if problem
```
agazso, line 128: I'm not sure I would add a comment to all invocation of the `pin` and `unpin` functions in tests. Instead it would be better to add a comment in the JSDoc comment of the functions.

AuHau: Well, I have added it because it is strange not to assert anything in tests.

agazso (approved):

nugaon (commented):
IMO It should be fixed in the Bee, do we have a ticket for this there?
and I also second Attila's suggestion.

AuHau:
> IMO It should be fixed in the Bee, do we have a ticket for this there?

Not sure what do you want to fix? I don't think there is anything wrong on Bee's side here. They use this structure to communicate for example errors as well. It is maybe a bit inconsistent as some endpoints return the data directly, while other endpoints follow this `BeeResponse` structure...

test/integration/bee-class.spec.ts
AuHau: To expand on the explanation. If somebody will be reading the tests and won't see any assertions he might wonder what are we testing? These comments are meant to explain how are we testing and not about how these methods work and hence I don't think JSDoc is a place where this should be put. That said, I will rephrase these comments to convey better the meaning.

nugaon:
> Not sure what do you want to fix? I don't think there is anything wrong on Bee's side here. They use this structure to communicate for example errors as well. It is maybe a bit inconsistent as some endpoints return the data directly, while other endpoints follow this `BeeResponse` structure...

It seems the pinning returns with `200` status code despite the operation failed, that I think is not correct. Otherwise the `axios` should throw an error I guess and the `safeAxios` would throw this error.

AuHau:
>It seems the pinning returns with 200 status code despite the operation failed, that I think is not correct. Otherwise the axios should throw an error I guess and the safeAxios would throw this error.

Wait, where do you see this? I am not aware of Bee returning `200` for failed operation...

PR #343, opened 2021-06-08
feat!: input validation

### LINKED ISSUE #202: Input validation of exposed funtions/methods

### LINKED ISSUE #248: Test input validation on Bee and BeeDebug classes
The methods exposed on the Bee and BeeDebug classes should have tests for the input validation. For now we are just going to test the incorrect inputs where we expect they assert.

_This comes as an outcome from a testing discussion on April 7th 2021_

AuHau:
Closes #202 and closes #248

### DIFF
--- a/src/bee-debug.ts
+++ b/src/bee-debug.ts
@@ -60,7 +60,9 @@ export class BeeDebug {
     return connectivity.getPeers(this.url)
   }
 
-  async removePeer(peer: string): Promise<RemovePeerResponse> {
+  async removePeer(peer: string | Address): Promise<RemovePeerResponse> {
+    assertAddress(peer)
+
     return connectivity.removePeer(this.url, peer)
   }
 
@@ -69,6 +71,8 @@ export class BeeDebug {
   }
 
   async pingPeer(peer: string): Promise<PingResponse> {
+    assertAddress(peer)
+
     return connectivity.pingPeer(this.url, peer)
   }
 
--- a/src/bee-debug.ts
+++ b/src/bee-debug.ts
@@ -70,7 +70,7 @@ export class BeeDebug {
     return connectivity.getTopology(this.url)
   }
 
-  async pingPeer(peer: string): Promise<PingResponse> {
+  async pingPeer(peer: string | Address): Promise<PingResponse> {
     assertAddress(peer)
 
     return connectivity.pingPeer(this.url, peer)
--- a/src/bee.ts
+++ b/src/bee.ts
@@ -20,7 +20,20 @@ import { createFeedManifest } from './modules/feed'
 import { assertBeeUrl, stripLastSlash } from './utils/url'
 import { EthAddress, makeEthAddress, makeHexEthAddress } from './utils/eth'
 import { wrapBytesWithHelpers } from './utils/bytes'
-import { assertBatchId, assertNonNegativeInteger, assertReference } from './utils/type'
+import {
+  assertAddressPrefix,
+  assertBatchId,
+  assertCollectionUploadOptions,
+  assertData,
+  assertFileData,
+  assertFileUploadOptions,
+  assertNonNegativeInteger,
+  assertPssMessageHandler,
+  assertPublicKey,
+  assertReference,
+  assertUploadOptions,
+  isTag,
+} from './utils/type'
 import { setJsonData, getJsonData } from './feed/json'
 import { makeCollectionFromFS, makeCollectionFromFileList } from './utils/collection'
 import { PostageBatchOptions, STAMPS_DEPTH_MAX, STAMPS_DEPTH_MIN } from './types'
@@ -50,6 +63,7 @@ import type {
   PostageBatch,
   BatchId,
 } from './types'
+// noinspection SuspiciousTypeOfGuard
 
 /**
  * The Bee class provides a way of interacting with the Bee APIs based on the provided url
@@ -95,6 +109,9 @@ export class Bee {
     options?: UploadOptions,
   ): Promise<Reference> {
     assertBatchId(postageBatchId)
+    assertData(data)
+
+    if (options) assertUploadOptions(options)
 
     return bytes.upload(this.url, data, postageBatchId, options)
   }
@@ -139,12 +156,19 @@ export class Bee {
     options?: FileUploadOptions,
   ): Promise<Reference> {
     assertBatchId(postageBatchId)
+    assertFileData(data)
+
+    if (options) assertFileUploadOptions(options)
+
+    if (name && typeof name !== 'string') {
+      throw new TypeError('name has to be string or undefined!')
+    }
 
     if (isFile(data)) {
       const fileData = await fileArrayBuffer(data)
-      const fileName = name || data.name
+      const fileName = name ?? data.name
       const contentType = data.type
-      const fileOptions = options !== undefined ? { contentType, ...options } : { contentType }
+      const fileOptions = options ? { contentType, ...options } : { contentType }
 
       return bzz.uploadFile(this.url, fileData, postageBatchId, fileName, fileOptions)
     } else {
@@ -193,6 +217,9 @@ export class Bee {
     options?: CollectionUploadOptions,
   ): Promise<Reference> {
     assertBatchId(postageBatchId)
+
+    if (options) assertCollectionUploadOptions(options)
+
     const data = await makeCollectionFromFileList(fileList)
 
     return bzz.uploadCollection(this.url, data, postageBatchId, options)
@@ -215,6 +242,8 @@ export class Bee {
     options?: CollectionUploadOptions,
   ): Promise<Reference> {
     assertBatchId(postageBatchId)
+
+    if (options) assertCollectionUploadOptions(options)
     const data = await makeCollectionFromFS(dir)
 
     return bzz.uploadCollection(this.url, data, postageBatchId, options)
@@ -233,6 +262,14 @@ export class Bee {
    * @param tagUid UID or tag object to be retrieved
    */
   async retrieveTag(tagUid: number | Tag): Promise<Tag> {
+    if (isTag(tagUid)) {
+      tagUid = tagUid.uid
+    } else if (typeof tagUid === 'number') {
+      assertNonNegativeInteger(tagUid, 'UID')
+    } else {
+      throw new TypeError('tagUid has to be either Tag or a number (UID)!')
+    }
+
     return tag.retrieveTag(this.url, tagUid)
   }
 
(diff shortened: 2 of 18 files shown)


src/bee.ts
```diff
@@ -46,10 +46,24 @@ import { createFeedManifest } from './modules/feed'
...
+  assertPssMessageHandler,
+  assertPublicKey,
+  assertReference,
+  assertUploadOptions,
+  isTag,
+} from './utils/type'
 import { setJsonData, getJsonData } from './feed/json'
 import { makeCollectionFromFS, makeCollectionFromFileList } from './utils/collection'
 import { PostageBatchOptions, STAMPS_DEPTH_MAX, STAMPS_DEPTH_MIN } from './types'
+// noinspection SuspiciousTypeOfGuard
```
agazso, line 66: What is this for? Maybe a comment would be good explaining why it is needed.

src/bee.ts
```diff
@@ -139,12 +156,19 @@ export class Bee {
...
+      throw new TypeError('name has to be string or undefined!')
+    }
 
     if (isFile(data)) {
       const fileData = await fileArrayBuffer(data)
-      const fileName = name || data.name
+      const fileName = name ?? data.name
       const contentType = data.type
-      const fileOptions = options !== undefined ? { contentType, ...options } : { contentType }
+      const fileOptions = options ? { contentType, ...options } : { contentType }
```
agazso, line 171: This would also work:
```suggestion
      const fileOptions = { contentType, ...options }
```

src/chunk/signer.ts
```diff
@@ -89,7 +89,7 @@ export function makePrivateKeySigner(privateKey: PrivateKeyBytes): Signer {
 }
 
 export function assertSigner(signer: unknown): asserts signer is Signer {
-  if (typeof signer !== 'object' || signer === null) {
+  if (typeof signer !== 'object' || signer === null || Array.isArray(signer)) {
```
agazso, line 92: Can you please explain the use-case when the `signer` is an array?

src/utils/collection.ts
```diff
@@ -29,6 +26,14 @@ export function assertCollection(data: unknown): asserts data is Collection<Uint
  * @param dir absolute path to the directory
  */
 export async function makeCollectionFromFS(dir: string): Promise<Collection<Uint8Array>> {
+  if (typeof dir !== 'string') {
+    throw new TypeError('dir has to be string!')
+  }
+
+  if (dir === '') {
+    throw new TypeError('dir most not be empty string!')
```
agazso, line 34: Typo:
```suggestion
    throw new TypeError('dir must not be empty string!')
```

AuHau: Thx!

src/chunk/signer.ts
AuHau: This is an assertion that Signer is **not** an array ;-) Because arrays are objects, so simple `typeof signer !== 'object'` won't catch that.

src/bee.ts
AuHau: Heh, true!

src/bee-debug.ts
```diff
@@ -69,6 +71,8 @@ export class BeeDebug {
   }
 
   async pingPeer(peer: string): Promise<PingResponse> {
```
agazso, line 73: The `peer` argument of `removePeer` has type `string | Address`, here it's only `string` although both are asserted as `Address`. For consistency it would be better to use the same types.

src/bee.ts
```diff
@@ -1,3 +1,6 @@
+// noinspection SuspiciousTypeOfGuard
+// Code quality instruction for IntelliJ IDEs
+
```
agazso, line 3: It looks like again your IDE is inserting stuff 😄

src/utils/type.ts
```diff
@@ -1,14 +1,41 @@
+// noinspection SuspiciousTypeOfGuard
+// Code quality instruction for IntelliJ IDEs
+
```
agazso, line 3: And again 😄

agazso (approved):
I think it would be better if you could change the TS project settings of your IDE and it would not be necessary to add the control comments in the committed file.

AuHau:
> I think it would be better if you could change the TS project settings of your IDE and it would not be necessary to add the control comments in the committed file.

Well, then can I remove the `.vscode` folder? 🤪

src/bee.ts
```diff
@@ -313,11 +349,23 @@ export class Bee {
...
-    recipient?: PublicKey,
+    recipient?: string | PublicKey,
   ): Promise<void> {
+    assertData(data)
     assertBatchId(postageBatchId)
+    assertAddressPrefix(target)
+
+    if (typeof topic !== 'string') {
+      throw new TypeError('topic has to be an string!')
+    }
```
nugaon, line 360: Here the topic should be a 32 lengthed hex string without prefix, shouldn't it?

src/bee.ts
```diff
@@ -329,6 +377,12 @@ export class Bee {
    * @returns Subscription to a given topic
    */
   pssSubscribe(topic: string, handler: PssMessageHandler): PssSubscription {
+    assertPssMessageHandler(handler)
+
+    if (typeof topic !== 'string') {
+      throw new TypeError('topic has to be an string!')
+    }
```
nugaon, line 384: Here the topic should be a 32 lengthed hex string without prefix, shouldn't it?

src/bee.ts
```diff
@@ -386,6 +440,14 @@ export class Bee {
    * @returns Message in byte array
    */
   async pssReceive(topic: string, timeoutMsec = 0): Promise<Data> {
+    if (typeof topic !== 'string') {
+      throw new TypeError('topic has to be an string!')
+    }
```
nugaon, line 445: Here the topic should be a 32 lengthed hex string without prefix, shouldn't it?

src/bee.ts
nugaon: In `topic.ts` you assert this in makeTopic with `assertBytes(topic, TOPIC_BYTES_LENGTH)` or you just create a hexstring with `makeHexString(topic, TOPIC_HEX_LENGTH)` if the `topic` is a `string`. why does it differ here?

src/bee.ts
AuHau: I don't think so. This is a PSS topic which I think is different from Feed's topic.

src/bee.ts
AuHau: This is PSS topic not Feed's topic.

src/bee.ts
AuHau: PSS not Feed.

src/utils/type.ts
```diff
@@ -39,9 +63,160 @@ export function assertReference(value: unknown): asserts value is Reference {
...
+  const options = value as UploadOptions
+
+  if (options.pin && typeof options.pin !== 'boolean') {
+    throw new TypeError(`options.pin property in ${name} has to be boolean or undefined!`)
+  }
+
+  if (options.encrypt && typeof options.encrypt !== 'boolean') {
+    throw new TypeError(`options.encrypt property in ${name} has to be boolean or undefined!`)
+  }
```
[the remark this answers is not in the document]
AuHau: No you are not right. I originally thought you are right but now realized that these options are optional, so `undefined` is a valid value so that is why there is the first part of the check.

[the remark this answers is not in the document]
nugaon: Screenshot from 2021-06-09 16-20-47

src/bee.ts
nugaon: In what points are different? and why we don't distinguish feed topic from pss topic then?

AuHau: Honestly, I am not sure, but this is already confusing on Bee's side. Feed topic is advertised as hex string as per [API reference](https://docs.ethswarm.org/api/#tag/Feed/paths/~1feeds~1{owner}~1{topic}/post) while PSS's topic is [advertised as generic string](https://docs.ethswarm.org/api/#tag/Postal-Service-for-Swarm/paths/~1pss~1send~1{topic}~1{targets}/post). 

I am all in to clear this up, but I don't think this is the time for that. I would start from Bee side to clarify what is expected, update API reference and then we can tackle that on bee-js side.

nugaon: that's why I'm interested, because in `swarm-cli` we handle the [PSS topics]() in the same way like the [Feed topics](https://github.com/ethersphere/swarm-cli/blob/b5189360d93ea76c4af8035aba1b85a61d53e9c4/src/command/feed/feed-command.ts#L30).

src/utils/type.ts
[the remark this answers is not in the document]
vojtechsimetka: @AuHau is right here. This is not what the condition is testing. The condition is testing

1. If the argument is set (thats the `if (options.encrypt)`)
2. Once it is verified the argument is set, it tests if it has correct type (`if(options.encrypt !== 'boolean')`)

If you would have just `if (options.encrypt !== 'boolean')` then for `options` object without the `encrypt` property it will throw which is not what we want.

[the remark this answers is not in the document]
AuHau: Yeah, agree with @vojtechsimetka here. That is the goal.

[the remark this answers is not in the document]
nugaon: > If you would have just `if (options.encrypt !== 'boolean')` then for `options` object without the `encrypt` property it will throw which is not what we want.

but I just demonstrated with my previous comment it doesn't throw any error... the typeof and undefined variable is `undefined`.

src/chunk/signer.ts
nugaon: we shouldn't support but I don't see why it is necessary to check, because it does not cause any failure

src/utils/type.ts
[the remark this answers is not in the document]
AuHau: Yes, but the problem here is not that `typeof undefined` would throw an error, but that `'undefined' !== 'boolean'` which then triggers the `if` and throw the 

```
throw new TypeError(`options.encrypt property in ${name} has to be boolean or undefined!`)
```

src/chunk/signer.ts
vojtechsimetka: I also don't think we need to support such array definition. That being said, the fact that we have such discussion indicates the code is confusing. Maybe this should be a different condition checking what the input **SHOULD** be and not what it **SHOULD NOT** be. So something like:
```ts
if (!(
    typeof singer === `object`
    typeof signer.sign === 'function' &&
    typeof signer.address === 'address' &&
    ...
    ))
```

src/utils/type.ts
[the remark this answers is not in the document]
vojtechsimetka: So to demonstrate this better:
```ts
options = undefined           // VALID
options = { encrypt: true }   // VALID

options = {}                  // throws
options = []                  // throws
options = { encrypt: 0 }      // throws
```

Your suggestion means that `options = undefined` throws as well.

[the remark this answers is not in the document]
nugaon: aaa it could be undefined as well? sorry I haven't read the text there. then all is okay. I thought the same proble happened as typeof name below.

src/bee.ts
agazso: I agree that this is confusing on the Bee side and I [opened an issue about it](https://github.com/ethersphere/bee/issues/2041). It seems that the way how we handle PSS topics in swarm-cli is incorrect at the moment.

PR #345, opened 2021-06-10
refactor!: string instead of bigint

### LINKED ISSUE #291: Accounting APIs return `number` and not `bigint` types In browsers
Tested in Brave, Chrome and Safari.

Here is an example code https://github.com/ethersphere/examples-js/tree/test/browser-bigint/acccounting-test

### LINKED ISSUE #335: Use only `bigint` for big amount numbers
Using `number | bigint` is a bad practice IMHO as when using `number` it might "randomly" (from user perspective) starts to throw errors once the safe limit is exceeded, which might or might not be checked by the developer that uses this library. Enforcing using `bigint` will decrease bit of DX, but I think would be more error-prone. What do you think? 
@vojtechsimetka

nugaon:
Bee started to return `string` instead of `number` from its API after [this PR](https://github.com/ethersphere/bee/pull/1735) was merged.
It yield to a refactoring within `bee-js` and grabbing the opportunity we decided to drop the `bigint` handling after we saw it rather makes complicated and cumbersome its usage for applications that use our library.

Other changes:
- make test postage stamp buying process faster
- merge jest matcher defnitions
- other refactors on the way

Closes #335, Closes #291


README.md
```diff
@@ -74,7 +74,7 @@ import { Bee } from "@ethersphere/bee-js"
 bee = new Bee("http://localhost:1633")
 
 // Be aware, this creates on-chain transactions that spend Eth and BZZ!
-const batchId = await bee.createPostageBatch(BigInt('100'), 17)
+const batchId = await bee.createPostageBatch('100'), 17)
```
agazso, line 77: Too many closing parentheses :)
```suggestion
const batchId = await bee.createPostageBatch('100', 17)
```

test/utils.ts
```diff
@@ -90,6 +77,19 @@ export function commonMatchers(): void {
             pass: false,
           }
     },
+    toBeNumberString(received) {
+      const message = () => `expected ${received} to be a number in a string type`
+
+      return /-?[0-9]+/g.test(received)
```
Cafe137, line 83: `/-?[0-9]+/g.test('not a number 1000')` evaluates `true`.

Please use `/^\-?[0-9_]+$/`

src/utils/type.ts
```diff
@@ -34,24 +35,24 @@ export function isUint8Array(obj: unknown): obj is Uint8Array {
   return obj instanceof Uint8Array
 }
 
-export function isInteger(value: unknown): value is number | bigint {
+export function isInteger(value: unknown): value is number | NumberString {
   return (
-    typeof value === 'bigint' ||
+    (typeof value === 'string' && /-?[0-9]+/g.test(value)) ||
```
Cafe137, line 40: Please use `/^\-?[0-9]+$/`

This tests for `true` for `'not a number 1000'`

src/utils/safe-axios.ts
```diff
@@ -64,39 +63,15 @@ function jsonBigTransformRequest(data: unknown, headers: Record<string, unknown>
...
-
-export async function safeAxios<T>(config: AxiosRequestConfig & { forceBigInt?: boolean }): Promise<AxiosResponse<T>> {
+export async function safeAxios<T>(config: AxiosRequestConfig): Promise<AxiosResponse<T>> {
   try {
     const response = await axios({
       maxContentLength: Infinity,
       maxBodyLength: Infinity,
-      transformRequest: jsonBigTransformRequest,
-      transformResponse: jsonBigTransformResponse(config.forceBigInt),
+      transformRequest: transformRequest,
```
Cafe137, line 74: Simply `transformRequest,`?

Cafe137 (approved):
LGTM

PR #358, opened 2021-06-22
ci: use bee-factor master and node 16

AuHau:


Cafe137 (approved):
Looks good, repo setting to be adjusted to remove 15

PR #361, opened 2021-06-25
feat: expose axios instance

Cafe137:


AuHau:
After discussion, we decided to use different approach as described in https://github.com/ethersphere/bee-js/issues/362

PR #365, opened 2021-07-06
feat: new tags endpoints

### LINKED ISSUE #353: Require `options.size` when uploading file with `Readable`
As observed in `swarm-cli`:

To upload a single file using the `uploadFile` function, we tried two things:

- Reading the data to memory with `readFileSync` and then passing that to upload
- Creating a `Readable` with `createReadStream` and then passing that to upload

After uploading, we query the `tag` to get the `total` amount of chunks, as well as the `processed` and `synced` status.

When using `readFileSync`, the `total` amount of chunks is correct. With `createReadStream`, it is always `2` or `3`, even when we expect thousands. 

---

To understand this `HTTP` / `axios` behaviour, I created a sandbox project with one server and two clients.

The server simply prints the incoming chunk sizes:

```
import { createServer } from 'http'

createServer((req, res) => {
    req.on('data', chunk => {
        console.log(chunk.toString().length)
    })
    req.on('end', () => {
        res.end('ok')
    })
}).listen(3333)
```

Then I create a ~15MB binary file to `POST`:

```
dd if=/dev/urandom count=1500 bs=10000 > data.bin
```

And `POST` it using `readFileSync`:

```
import axios from 'axios'
import { readFileSync } from 'fs'

axios.post('http://localhost:3333', readFileSync('data.bin'), { maxContentLength: Infinity, maxBodyLength: Infinity })
```

The chunk sizes look very similar:

```
62027
62034
62045
61993
62191
62218
62184
61996
62177
```

Now with `createReadStream`:

```
import axios from 'axios'
import { createReadStream } from 'fs'

axios.post('http://localhost:3333', createReadStream('data.bin'), { maxContentLength: Infinity, maxBodyLength: Infinity })
```

The chunk sizes vary:

```
62018
203
61822
212
61825
218
61769
225
61955
237
61972
251
```

The assumption is that with streaming, a different `Transfer-Encoding` is used, probably `chunked`, which Bee does not handle correctly.

My other theory is that the `Content-Length` header is missing or is wrong, and Bee uses that to determine total amount of chunks.

### LINKED ISSUE #363: Missing Tag endpoints
We are currently missing support for:

 - Listing all Tags
 - Deletion of Tags
 - Updating `total` field for streamed data

https://docs.ethswarm.org/api/#tag/Tag/paths/~1tags/get

AuHau:
Closes #363 and closes #353

### DIFF
--- a/src/bee.ts
+++ b/src/bee.ts
@@ -22,6 +22,7 @@ import { EthAddress, makeEthAddress, makeHexEthAddress } from './utils/eth'
 import { wrapBytesWithHelpers } from './utils/bytes'
 import {
   assertAddressPrefix,
+  assertAllTagsOptions,
   assertBatchId,
   assertBoolean,
   assertCollectionUploadOptions,
@@ -34,19 +35,11 @@ import {
   assertReference,
   assertUploadOptions,
   isReadable,
-  isTag,
+  makeTagUid,
 } from './utils/type'
 import { setJsonData, getJsonData } from './feed/json'
 import { makeCollectionFromFS, makeCollectionFromFileList } from './utils/collection'
-import {
-  AllTagsOptions,
-  NumberString,
-  PostageBatchOptions,
-  STAMPS_DEPTH_MAX,
-  STAMPS_DEPTH_MIN,
-  TAGS_LIMIT_MAX,
-  TAGS_LIMIT_MIN,
-} from './types'
+import { AllTagsOptions, NumberString, PostageBatchOptions, STAMPS_DEPTH_MAX, STAMPS_DEPTH_MIN } from './types'
 
 import type {
   Tag,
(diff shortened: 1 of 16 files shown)


src/bee.ts
```diff
@@ -283,20 +298,63 @@ export class Bee {
...
+   * @throws TypeError if limit or offset are not numbers or undefined
+   * @throws BeeArgumentError if limit or offset have invalid options
+   *
+   * @see [Bee docs - Syncing / Tags](https://docs.ethswarm.org/docs/access-the-swarm/syncing)
+   * @see [Bee API reference - `GET /tags`](https://docs.ethswarm.org/api/#tag/Tag/paths/~1tags/get)
+   */
+  async getAllTags(options?: AllTagsOptions): Promise<Tag[]> {
+    if (options !== undefined && (typeof options !== 'object' || options === null || Array.isArray(options))) {
+      throw new TypeError('options has to be an object or undefined!')
+    }
```
Cafe137, line 324: I think I have seen this object check before. So I was thinking maybe extract it? 🙂

I also run into this scenario several times and ended up writing these utilities:

```
const isObject = value => value !== null && typeof value === 'object'

const isStrictlyObject = value => isObject(value) && !Array.isArray(value)
```

Then it can become `options !== undefined && !isStrictlyObject(options)` which is a bit more readable.

src/bee.ts
```diff
@@ -310,6 +368,60 @@ export class Bee {
...
+  async updateTag(tagUid: number | Tag, reference: Reference | string): Promise<void> {
+    assertReference(reference)
+
+    if (isTag(tagUid)) {
+      tagUid = tagUid.uid
+    } else if (typeof tagUid === 'number') {
+      assertNonNegativeInteger(tagUid, 'UID')
+    } else {
+      throw new TypeError('tagUid has to be either Tag or a number (UID)!')
+    }
```
Cafe137, line 420: I think this is a duplicate block, and could be extracted to a helper function which gets `uid` from `number | Tag` or throws

Cafe137 (approved):
I see a few opportunities to extract helper functions (see my comments), but otherwise LGTM 🙂

PR #367, opened 2021-07-15
feat: hooks system

### LINKED ISSUE #362: Hooks system for requests and responses
For debugging purposes, it might be interesting for the users to see what exact requests the `bee-js` are sent out. 

The idea is to introduce a hook system where users can register their hooks and all the requests and responses are pass to.

AuHau:
Closes #362


src/utils/hooks.ts
```diff
@@ -0,0 +1,60 @@
...
+import { BeeRequest, BeeResponse, HttpMethod } from '../types'
+
+type HookCallback<V> = (value: V) => V | Promise<V>
+
+function wrapRequest(request: AxiosRequestConfig): BeeRequest {
+  let headers = request.headers
+
+  if (headers.common && headers.get && headers.delete && headers.post && headers.patch) {
+    headers = Object.assign({}, request.headers.common ?? {}, request.headers[request.method as HttpMethod] ?? {})
+  }
```
nugaon, line 11: I don't really get this part.

- At the condition the `get`, `delete`, `post`, `patch` and `common` have to be included in the `header`. Maybe those should be checked with `||`?
- why the condition is necessary if you coalesce check all objects? 
- Also, why does it overwrite the initial `header`?

I suggest something like this instead:

```suggestion
  const headers = Object.assign(request.headers ?? {}, request.headers.common ?? {}, request.headers[request.method as HttpMethod] ?? {})
```

src/types/index.ts
```diff
@@ -246,6 +246,24 @@ export interface ReferenceResponse {
   reference: Reference
 }
 
+export type HttpMethod = 'get' | 'GET' | 'delete' | 'DELETE' | 'post' | 'POST' | 'patch' | 'PATCH'
```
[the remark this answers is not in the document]
AuHau: Not to my knowledge. Only `PATCH`.

src/utils/hooks.ts
AuHau: Yeah, I could have added a comment about this.

This function is used for filtering of properties for the request's object and for the response object as well which has `config` property that can be used to get info about the originating request.

There is inconsistency in how the headers are formatted there. 

For requests object, it has this structure that it uses to configure common and specific method's headers:
```js
{
	common: {...},
	get: {...},
	post: {...},
	delete: {...},
	patch: {...}
}
```

so to get really the final headers list I have to join `common` and whatever the method the request is making.

For the response object, this is not the case and it contains already the final headers object. 

So I am doing this joining only when I detect that the object has this configuration format.

nugaon: ah okay, then it is strange indeed, but it's okay.

nugaon (approved):

PR #370, opened 2021-07-21
fix: provide all request headers

AuHau:
I got the wrong idea about the axios's headers object structure which lead to no headers reported for the request hook.


Cafe137 (approved):
I tried with `swarm-cli` and I see all expected headers, thanks!

PR #377, opened 2021-07-28
test: improve postage creation

AuHau:
Now you can configure env. variable for the postage stamps so they are not created on every test run.

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -71,31 +71,46 @@ export default async (): Promise<Config.InitialOptions> => {
     if (!process.env.BEE_POSTAGE || !process.env.BEE_PEER_POSTAGE) {
       console.log('Creating postage stamps...')
 
-      const stampsPromises: Promise<BatchId>[] = []
-      const stampsEnv = []
+      const stampsOrder: { url: string; env: string }[] = []
 
       if (!process.env.BEE_POSTAGE) {
-        stampsPromises.push(createPostageBatch(beeUrl, '1', 20))
-        stampsEnv.push('BEE_POSTAGE')
+        stampsOrder.push({ url: beeUrl, env: 'BEE_POSTAGE' })
       }
 
       if (!process.env.BEE_PEER_POSTAGE) {
-        stampsPromises.push(createPostageBatch(beePeerUrl, '1', 20))
-        stampsEnv.push('BEE_PEER_POSTAGE')
+        stampsOrder.push({ url: beePeerUrl, env: 'BEE_PEER_POSTAGE' })
       }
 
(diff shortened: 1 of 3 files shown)


jest.config.ts
```diff
@@ -36,19 +37,66 @@ export async function getBrowserPathMapping(): Promise<{ [aliasNodeReference: st
...
-    console.log('Peer stamp: ', process.env.BEE_PEER_POSTAGE)
-    // sleep for 11 seconds (10 blocks with ganache block time = 1s)
-    // needed for postage batches to become usable
-    // FIXME: sleep should be imported for this, but then we fail with
-    //        Could not find a declaration file for module 'tar-js'
-    await new Promise<void>(resolve => setTimeout(() => resolve(), 11_000))
+
+    if (process.env.BEE_POSTAGE) {
+      try {
+        if (!(await getPostageBatch(beeUrl, process.env.BEE_POSTAGE as BatchId)).usable) {
```
agazso, line 45: As far as I know the returned postage stamp is not usable for seconds after it was succesfully created.
See https://github.com/ethersphere/swarm-cli/pull/212/files

AuHau: Well, this is up to the user to handle. If he is too fast in creation of the stamps and running the tests... The idea here is that you run the tests on a fresh environment which creates the stamps and then you can configure it as env. variable. Then the stamps are already valid...

agazso: Ok, got it

jest.config.ts
```diff
@@ -36,19 +37,66 @@ export async function getBrowserPathMapping(): Promise<{ [aliasNodeReference: st
...
+      for (let i = 0; i < stamps.length; i++) {
+        process.env[stampsEnv[i]] = stamps[i]
+        console.log(`${stampsEnv[i]}: ${stamps[i]}`)
+      }
+
+      // sleep for 11 seconds (10 blocks with ganache block time = 1s)
+      // needed for postage batches to become usable
+      // FIXME: sleep should be imported for this, but then we fail with
+      //        Could not find a declaration file for module 'tar-js'
+      await new Promise<void>(resolve => setTimeout(() => resolve(), 11_000))
```
agazso, line 98: Maybe this could be replaced with a check for the postage stamp to be `usable` instead of just a timer for more deterministic tests.

AuHau: You mean in cycle querying until `usable = true`?

agazso: Yes. In my experience one of the biggest source of flaky tests are sleeps in the code which are based on assumptions that do not hold over time.

agazso (approved):

Issue #250, opened 2021-04-07
Use shorter prefixes in PSS tests to reduce test duration


agazso:
PSS test duration depends on calculating the mined trojan chunk data. With the default settings it can take as much as 120 seconds. The idea is to use shorter prefixes to reduce the time it takes to run the PSS tests.

AuHau:
Bump.

PR #384, opened 2021-08-02
fix!: pss target length verification

### LINKED ISSUE #380: PSS Send: target is truncated to 4 characters
https://github.com/ethersphere/bee-js/blob/e7e81fc4e9bfe75760392dfe46766ef846d6b5f5/src/modules/pss.ts#L31

AuHau:
Closes #380 

Most of the validation was already in place from my earlier validation work, I was just asserting wrong values 😅

### Breaking changes
 
 - `bee.pssSend()` now throws error if the specified target exceeds maximal value. Use `Utils.makeMaxTarget()` that will give you the max target that Bee accepts.


Cafe137 (approved):
Great, thanks :)

AuHau:
The only downside about this fix is that it is breaking 😒 But honestly the other work (directory streaming and fetch refactor) are also breaking so I guess we should just go with 2.0 release.

agazso (commented):
Using `makeMaxTarget` in the test may make the PSS tests to run longer and use more CPU than necessary. In general for testing PSS it would be enough to assume the shortest possible target length.

AuHau:
You are right, but that is not the scope of this PR. I was only reproducing the current state which was passing the whole address which then got truncated with the original `slice()`. 

I have created an issue about this improvement and I will work on it in the future.

agazso:
> You are right, but that is not the scope of this PR. I was only reproducing the current state which was passing the whole address which then got truncated with the original `slice()`.

Okay, I see.

> I have created an issue about this improvement and I will work on it in the future.

I also [opened one](https://github.com/ethersphere/bee-js/issues/250) long time ago, you may want to close either as duplicate.

AuHau:
Thanks for the info!

src/utils/pss.ts
```diff
@@ -0,0 +1,16 @@
+import { AddressPrefix, PSS_TARGET_HEX_LENGTH_MAX } from '../types'
+
+/**
+ * Utility function that for given strings/reference takes the most specific
+ * target that Bee node will except.
+ *
+ * @param target
```
nugaon, line 7: is the target prefixed with `0x` or not?

src/utils/pss.ts
```diff
@@ -0,0 +1,16 @@
+import { AddressPrefix, PSS_TARGET_HEX_LENGTH_MAX } from '../types'
+
+/**
+ * Utility function that for given strings/reference takes the most specific
+ * target that Bee node will except.
```
nugaon, line 5: does not the bee accept any length of target? then why we allow to pass any length address?

src/utils/pss.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * @param target
+ * @see [Bee docs - PSS](https://docs.ethswarm.org/docs/dapps-on-swarm/pss)
+ */
+export function makeMaxTarget(target: string): AddressPrefix {
```
nugaon, line 10: I would rename it `makeOptimalTargetPrefix`

nugaon (commented):
I don't completely get the API allows any length of target, but the BeeJS restricts it to 4-length hex string. Because we don't expose/support the direct use of modules as we discussed, by design you even cannot get around this limitation in this form, meanwhile you bring a utility function usage to the workflow. it seems more unconvenient for me with zero yield compare to the previous version.

AuHau:
@nugaon Bee API also has the same restrictions. It will return Bad Request for target exceeding the limit. See https://github.com/ethersphere/bee/blob/master/pkg/api/pss.go#L29 (btw. there the limit is 3 bytes = 6 hex chars, but this was bumped only after Bee 1.1.0 release, so it is not propagated to Bee-js for now and I will change it once release cycle will begin).

src/utils/pss.ts
AuHau: I disagree with the rename. This utility function does not make optimal target, but returns the most specific target that Bee will accept, which is not the same. Only the user can decide what is optimal for his use-case as more concrete target sacrifice performance for more "targetness"...

src/utils/pss.ts
AuHau: No. You might got this impression from Bee-js usage of PSS before this patch, which silently sliced the target to max. allowed length without user's knowledge. This patch impose correct validation and tools for the user to have predictable behavior.

src/utils/pss.ts
AuHau: No. It is Swarm address. Only blockchain hashes are prefixed.

nugaon: I mean could you make a description about it next to the param please?

src/utils/pss.ts
nugaon: if the bee does not accept higher accuracy target then this name is fine.

src/utils/pss.ts
AuHau:
```suggestion
 * @param target is a non-prefixed hex string Bee address
```

nugaon (approved):
as everybody is satisfied with this then we can roll it out, though from my side a jsdoc about this splitting would have been enough

PR #395, opened 2021-08-18
refactor!: utility module flatting and limiting

### LINKED ISSUE #385: "Over-exposed" utility functions
Looking on what is exposed as part of the `Utils` namespace, I think there is a lot of parts that are purely internal and I am not sure why user should have access to it. Especially as it extends the public API that is then tightened to SemVer.

```ts
export * as Collections from './collection' // <- should expose only the helpers the @Cafe137 created
export * as Bytes from './bytes' // <- also not sure if it is needed
export * as Data from './data' // <- I would definitely remove this one
export * as File from './file' // <- This will be removed in the directory streaming PR
export * as Hex from './hex' // <- I guess this one makes sense
export * as Tar from './tar' // <- I would definitely remove this one
export * as Eth from './eth' // <- This one makes sense
export * as Hooks from './hooks' // <- This one could be only deprecated with the Fetch PR, but if we gonna have breaking release we might remove it as well.
export * as Pss from './pss'
export { setDefaultHeaders } from './safe-axios'
export { keccak256Hash } from './hash'
```

AuHau:
Closes #385

### DIFF
--- a/src/utils/expose.ts
+++ b/src/utils/expose.ts
@@ -1,9 +1,47 @@
-export * as Collections from './collection'
-export * as Bytes from './bytes'
-export * as Data from './data'
-export * as File from './file'
-export * as Hex from './hex'
-export * as Tar from './tar'
-export * as Eth from './eth'
-export * as Stream from './stream'
+export { getCollectionSize, getFolderSize } from './collection'
+
+export {
+  Bytes,
+  FlexBytes,
+  isBytes,
+  assertBytes,
+  isFlexBytes,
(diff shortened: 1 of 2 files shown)


vojtechsimetka (approved):
Yes please!

agazso (approved):

nugaon (approved):
after this merge we have to think about adding every new utility function to this list.

it can be merged from my side, but anyway, we will have `swarm-js-utils` soon : )

PR #399, opened 2021-08-23
feat: pending transactions management

### LINKED ISSUE #396: Cancel transaction support
https://github.com/ethersphere/bee/pull/2212/

AuHau:
Closes #396

### DIFF
--- a/src/bee-debug.ts
+++ b/src/bee-debug.ts
@@ -3,6 +3,7 @@ import * as balance from './modules/debug/balance'
 import * as chequebook from './modules/debug/chequebook'
 import * as settlements from './modules/debug/settlements'
 import * as status from './modules/debug/status'
+import * as transactions from './modules/debug/transactions'
 import * as states from './modules/debug/states'
 import type {
   Address,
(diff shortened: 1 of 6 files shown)


src/bee-debug.ts
```diff
@@ -380,4 +390,49 @@ export class BeeDebug {
...
+  async getPendingTransaction(transactionHash: TransactionHash | string): Promise<TransactionInfo> {
+    assertTransactionHash(transactionHash)
+
+    return transactions.getTransaction(this.url, transactionHash)
+  }
+
+  /**
+   * Rebroadcast already created transaction
+   * @param transactionHash
+   */
```
Cafe137, line 414: I think the JSDoc should explain what rebroadcast is good for, when to use it.

The function is called `rebroadcastPending` but the JSDoc says _Rebroadcast already created_, not sure if this is intentional, but it is a bit confusing to me. 🙂

AuHau: Yeah agree. I'm just not sure exactly what is it about so choose not to expand on that 😅

Maybe @vojtechsimetka could you please help me out with what should be written here?

nugaon: For me the PR's code is good, but this part also little bit vague for me what this function does.

vojtechsimetka: From the Ethereum perspective this _Rebroadcasts already created transaction_ also called _pending transaction_ (as it is a transaction that has been before send to the network but is not yet included in any block). Not sure what is the confusion here :) . In the other function you have comment _Rebroadcast existing transaction_ which imho is also OK.

vojtechsimetka (approved):

nugaon (approved):

PR #408, opened 2021-08-31
feat!: return tags for uploads

### LINKED ISSUE #366: Tags are automatically created - should we automatically return them?
All the upload endpoints upon successful upload automatically create a Tag that is returned in the response header `Swarm-Tag`. 

Currently, we return only the reference but should we return the object instead with both values?

AuHau:
Closes #366 

### Breaking changes:

 - Changes the returned type to `UploadResult` interface for all uploading methods like: `uploadData()`, `uploadFile()`, `uploadFilesFromDirectory()`, `uploadFiles()`

### Other changes:

 - Introduces `uploadCollection()` method


test/integration/bee-class.spec.ts
```diff
@@ -175,9 +176,37 @@ describe('Bee class', () => {
...
+      ]
+
+      const result = await bee.uploadCollection(getPostageBatch(), directoryStructure)
+      const file = await bee.downloadFile(result.reference, directoryStructure[0].path)
+
+      expect(file.name).toEqual(directoryStructure[0].path)
+      expect(file.data.text()).toEqual('hello-world')
+    })
+
+    it('should upload browser files', async () => {
```
nugaon, line 199: I don't completely get why it is placed here. we have a real browser environment for testing.

src/utils/http.ts
```diff
@@ -97,6 +97,7 @@ export async function http<T>(ky: Ky, config: HttpOptions): Promise<KyResponse<T
 
     const response = (await ky(path, {
       ...kyConfig,
+      mode: 'cors',
```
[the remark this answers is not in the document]
AuHau: Thanks for the catch! Yeah, this is a leftover from the times when I was trying to debug why the upload in browsers did not work. I will remove it.
```suggestion
```

test/integration/bee-class.spec.ts
AuHau: Well, this test makes sure that this API works in Node as well...

nugaon: But if it works in browser environment then why is it necessary to test it on node side? I mean just because of it we need an additional devDep. what use-cases you see to use the upload like this on node-side?

AuHau: Honestly, this test comes from #372 and I don't really remember why I have done it like this anymore. I will move it to browser tests for now and will see if I will find the reason again 😅

AuHau: I see, it is already in the browser tests. 

Btw. this is again related to isomorphic testing. The fact that we expose some API should be tested in all environments because only then we can be sure it works. I could imagine a case where we could do some checks like `file instanceof File` and it would work in the browser but would not work in Nodejs. But our API is pronouncing support of `File` everywhere, so somebody could because of some weird reasons try that in Node and it would not work...

nugaon:
Please remove the `@web-std/file` from the devDeps as well.

nugaon (approved):

PR #414, opened 2021-09-14
refactor: use global setup for stamps creation in tests

### LINKED ISSUE #391: Refactor the Postage Stamps buy process out of `jest.config` into `globalSetup` module
Having a async setup in config is quite big hack. Jest supports `globalSetup`/`globalTeardown` that could be exactly used for this purpose!

https://stackoverflow.com/questions/48318230/configure-jest-global-tests-setup-with-ts-file-typescript

AuHau:
Closes #391

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 /*
  * For a detailed explanation regarding each configuration property and type check, visit:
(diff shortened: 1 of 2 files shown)


nugaon (approved):
yeah, this is tidier.

We should apply the bee setup like this at our other (new) repos as well.

PR #415, opened 2021-09-20
test: shorten pss test target

### LINKED ISSUE #250: Use shorter prefixes in PSS tests to reduce test duration

AuHau:
Closes #250


nugaon (commented):
You have left out the pss integration test file where you also send pss messages.

PR #440, opened 2021-11-03
fix(build): sourcemap

nugaon:
> If the default webpack minimizer has been overridden (such as to customise the terser-webpack-plugin options), make sure to configure its replacement with `sourceMap: true` to enable SourceMap support.

_Source: https://webpack.js.org/configuration/devtool/_


AuHau (approved):
Lgtm

PR #442, opened 2021-11-03
ci: enable dependabot

AuHau:


nugaon:
I don't completely get this CI action. Why is it not appropriate to allow only the `Dependabot security updates`?

vojtechsimetka (approved):

PR #444, opened 2021-11-05
fix: readable-stream should be dependency

AuHau:


nugaon:
Wasn't it the problem that `readable-stream` itself was in the `devdDpendencies` list?

nugaon (approved):

PR #459, opened 2021-11-23
feat: openapi versions in health

### LINKED ISSUE #454: Add OpenAPI versions to types
https://github.com/ethersphere/bee/pull/2638

AuHau:
Closes #454

### DIFF
--- a/src/types/debug.ts
+++ b/src/types/debug.ts
@@ -144,6 +144,8 @@ export interface BalanceResponse {
 export interface Health {
   status: 'ok'
   version: string
+  apiVersion: string
+  debugApiVersion: string
 }
 
 export interface RemovePeerResponse {


nugaon (commented):
I think the `isSupportedVersion` logic should be changed of the `BeeDebug` class.

AuHau:
@nugaon how?

nugaon:
like it checks the supported API version instead of the bee version (which is hard-coded in the package.json's `engines.bee`).

AuHau:
Well, it would complicate a lot of things at the moment. We would have to include also the OpenAPI versions, which would also break the Bee upgrade workflow. I am not against it, but please create an issue for this and we can plan it for future releases.

nugaon (approved):

PR #461, opened 2021-11-24
ci: fix bee update flow

AuHau:


.github/workflows/update_bee.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
           fi
 
           echo "IMAGE_VERSION=$image_version" >> $GITHUB_ENV
-          echo "BEE_VERSION=${$image_version/-stateful}" >> $GITHUB_ENV
+          echo "BEE_VERSION=${image_version/-stateful}" >> $GITHUB_ENV
```
nugaon, line 29: wouldn't it be more safe and readable?
```suggestion
          echo "BEE_VERSION=${image_version}-stateful" >> $GITHUB_ENV
```

AuHau: Thanks but no thanks... This is Bash's string find&replace, which removes `-stateful` part out of the Bee version that is passed from Bee-factory.

nugaon: ah, okay

nugaon (approved):

PR #477, opened 2021-12-08
feat: exposing chunk functionality

### LINKED ISSUE #476: Readable support for `/chunks`
Because of some inconsistencies in the test, I have not merged as part of #472 support for fetching chunk with readable. This should be more investigated and fixed.

AuHau:
Closes #476


AuHau:
We have decided to not implement readable support for Chunk as it does not really make sense as it is at most 4K of data...

PR #479, opened 2021-12-08
feat: node endpoint support

### LINKED ISSUE #474: Node endpoint support
https://github.com/ethersphere/bee/pull/2649

—

Created via [Raycast](https://www.raycast.com?ref=signatureGithub)

AuHau:
Closes #474


src/modules/debug/status.ts
```diff
@@ -1,19 +1,38 @@
...
+
+  return response.data
+}
+
+/**
+ * Get mode of Bee node
+ *
+ * @param ky Ky debug instance
+ */
+export async function getModes(ky: Ky): Promise<Modes> {
```
nugaon, line 32: The URL later potentially can serve other data, (e.g. https://github.com/ethersphere/bee/pull/2649#discussion_r758530593 ) so I would keep the interface naming close to the actual endpoint's name itself, in this case `getNode` (or maybe `getNodeInfo`).

test/integration/bee-debug-class.spec.ts
```diff
@@ -111,4 +111,15 @@ describe('Bee Debug class', () => {
       await expect(beeDebug.createPostageBatch('-1', 17)).rejects.toThrowError(BeeArgumentError)
     })
   })
+
+  describe('modes', () => {
+    it('should return modes', async () => {
+      expect(await beeDebug.getModes()).toEqual(
+        expect.objectContaining({
+          beeMode: expect.any(String),
```
nugaon, line 119: I would check whether `beeMode` is `light`, `full` or `dev`, otherwise throw an error.

src/modules/debug/status.ts
AuHau: Ok, I originally didn't like the `getNode` as I think it does not give you much context, but I like `getNodeInfo`. I will change it.

src/modules/debug/status.ts
```diff
@@ -1,33 +1,52 @@
...
+
+/**
+ * Get information about Bee node
+ *
+ * @param ky Ky debug instance
+ */
+export async function getNodeInfo(ky: Ky): Promise<NodesInfo> {
+  const response = await http<NodesInfo>(ky, {
+    method: 'get',
+    path: MODES_URL,
```
nugaon, line 35: use `NODE_INFO_URL` instead

nugaon (approved):

PR #508, opened 2022-01-11
fix: handling of non-json responses when json is expected

AuHau:


test/unit/utils/http.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+import { BeeNotAJsonError, BeeResponseError } from '../../../src'
+
+describe('http', () => {
+  it('should handle non-json response for 200', async () => {
+    const HTML_RESPONSE = `<html><body><h1>Some error!</h1></body></html>`
+
+    nock(MOCK_SERVER_URL).get('/endpoint').reply(200, HTML_RESPONSE)
+    const kyInst = ky.create({ prefixUrl: MOCK_SERVER_URL })
+
+    await expect(http(kyInst, { path: 'endpoint', responseType: 'json', method: 'get' })).rejects.toThrow(
```
vojtechsimetka, line 14: I guess this is enough of a test. I was wondering if it would be even better to mock some existing Bee endpoint and check if the instance of `Bee` class throws this error, right? E.g.:

```ts
nock(MOCK_SERVER_URL).get('/bzz/fa421...78a93').reply(200, HTML_RESPONSE)
new Bee('...')
await expect(bee. downloadFile('fa421...78a93')).rejects.toThrow(BeeNotAJsonError)
```

vojtechsimetka (approved):

test/unit/utils/http.spec.ts
AuHau: These are unit tests so I would keep it as is.

vojtechsimetka: It could still be unit tests but make sure it not only works with a KY instance but encapsulated in the Bee class :)

PR #517, opened 2022-01-24
feat: exposing request and response configuration on BeeResponseError

### LINKED ISSUE #462: Add request and response objects to BeeResponseError
I am working with a Node.js based command-line tool that uses bee-js and when there is a problem, the only information I can get from the `BeeResponseError` object is the error message and the HTTP status code. However sometimes it is not enough information to see what went wrong, especially when working with a Bee node where I have no access to the error logs.

It would be good if the `BeeResponseError` would also include information about the request and the response that was made, so that there is more context when debugging an issue.

AuHau:
Closes #462


vojtechsimetka (approved):
LGTM

PR #533, opened 2022-02-03
feat: hybrid esm build

AuHau:
This PR introduces a hybrid build supporting both CommonJS and ESM. It uses `tsc` for compilation and `webpack` was scoped down only for browser bundle UMD builds. 

As in Node importing JSON modules is [experimental](https://nodejs.org/api/esm.html#esm_json_modules) and hidden behind a flag, I have decided to drop getting the Bee version from `package.json` and copied the information directly to `/src/modules/debug/status.ts` file where it is used and exposed to the codebase. I have modified the `update_bee` workflow which should also replace this version upon updating the Bee version. I have not removed the `engine.Bee` property in `package.json` (and its updating mechanism in `update_bee` workflow) as I think it is right place to have this information as well and might be beneficial for some other tools like our discussed refactor of Bee Factory


build-fixup
```diff
@@ -0,0 +1,21 @@
...
+
+cat >dist/mjs/package.json <<!EOF
+{
+    "type": "module",
+    "browser": {
+      "./utils/data.js": "./utils/data.browser.js",
+      "./utils/collection.node.js": "./utils/collection.browser.js"
+    }
+}
+!EOF
```
vojtechsimetka, line 21: Could you elaborate what this does and why it is needed please? Maybe there should be some comment about it.

src/utils/http.ts
```diff
@@ -3,14 +3,13 @@ import type { BeeRequest, BeeResponse, HookCallback, HttpMethod, Ky } from '../t
...
 import { deepMerge } from './merge'
-import { version as beeJsVersion } from '../../package.json'
 import { isObject, isStrictlyObject } from './type'
 import { KyRequestOptions } from '../types'
 
 const DEFAULT_KY_CONFIG: KyOptions = {
   headers: {
     accept: 'application/json, text/plain, */*',
-    'user-agent': `bee-js/${beeJsVersion}`,
+    'user-agent': `bee-js`,
```
vojtechsimetka, line 12: Shouldn't we keep this?

vojtechsimetka (commented):
Looks good but I would like to properly test this. Did you check if it works with the examples repo?

build-fixup
AuHau: Done, let me know if it needs more clarification or it is alright like this.

src/utils/http.ts
AuHau: Well ideally yes, but the version was imported from `package.json` and I needed to remove all JSON imports. I think it is not critical thing to keep in so decided to remove it instead looking for workarounds like with the `SUPPORTED_BEE_VERSION_EXACT` if it turns out that somebody is depending on it we can think of alternatives.

AuHau:
> Did you check if it works with the examples repo?

Yes, I have NPM linked this branch with `upload-progress` and debugged the build with it.
I had to tweak the `packageJson.browser` so we can produce build that does not relly on any Node's data structures, so now it is possible to build with webpack without the need for any specification of the nodeJs internal mockings etc.

vojtechsimetka (approved):
LGTM, tested with `bee-dashboard` and `swarm-cli`.

PR #534, opened 2022-02-03
feat: add owner property to SOCReader interface

### LINKED ISSUE #532: It would be good to have the `owner` exposed on SOCReader
When using `SOCWriter` sometimes I need the `owner` address. For example when passing it to a `SOCReader` communicated in a url.

Of course it is possible to get this by using an external library that deals with Ethereum keys, however it is inconvenient to do so.

It would be good to expose the `owner` property on the `SOCReader` interface (similarly how it is done in the `FeedReader` interface). Then it would be automatically available on the `SOCWriter` as well.

agazso:
Fixes #532

### DIFF
--- a/src/bee.ts
+++ b/src/bee.ts
@@ -970,6 +970,7 @@ export class Bee {
     const canonicalOwner = makeEthAddress(ownerAddress)
 
     return {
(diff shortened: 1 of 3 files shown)


AuHau (approved):
LGTM

PR #550, opened 2022-02-10
fix: use default import styles for esm compatibility

AuHau:
@Cafe137 identified that when used the new 3.3.0 released in Node and ESM it will fail as:

```
> gateway-watch@1.0.0 start
> node --experimental-modules --es-module-specifier-resolution=node index.js

file:///Users/aron/Code/gateway-watch/node_modules/@ethersphere/bee-js/dist/mjs/utils/hash.js:1
import { keccak256 } from 'js-sha3';
         ^^^^^^^^^
SyntaxError: Named export 'keccak256' not found. The requested module 'js-sha3' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'js-sha3';
const { keccak256 } = pkg;
```

This should fix this.

### DIFF
--- a/src/chunk/signer.ts
+++ b/src/chunk/signer.ts
@@ -1,4 +1,9 @@
-import { ec, curve } from 'elliptic'
+import { curve } from 'elliptic'
+
+// For ESM compatibility
+import pkg from 'elliptic'
+const { ec } = pkg
+
 import { BeeError } from '../utils/error'
 import { Bytes, isBytes, assertBytes, wrapBytesWithHelpers } from '../utils/bytes'
 import { keccak256Hash } from '../utils/hash'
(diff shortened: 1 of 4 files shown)


vojtechsimetka (approved):
Did you test it with `npm link`? Also shouldn't these have `// FIXME` tag when the `js-sha3` gets updated to ESM?

src/chunk/signer.ts
```diff
@@ -1,4 +1,9 @@
-import { ec, curve } from 'elliptic'
+import { curve } from 'elliptic'
+
+// For ESM compatibility
+import pkg from 'elliptic'
+const { ec } = pkg
+
```
Cafe137, line 6: Shouldn't this be `const { curve, ec } = pkg`?

My assumption is that `import { curve } from 'elliptic'` will raise an error too 🙁

AuHau:
Well, I doubt `js-sha3` will be updated to ESM. The last commit was 4 years ago and it seems pretty unmaintained. But it is not a problem for us as you can import CommonJS packages to ESM, the problem is vice-versa.

PR #556, opened 2022-02-14
fix: making readable-stream optional

AuHau:
This PR makes the `readable-stream` dependency optional. It is needed only when `Utils.readableWebToNode()` function is used in browsers.

I have debugged this and verified with new example `browser-stream` that is placed here: https://github.com/ethersphere/examples-js/pull/26 and will be merged once the `bee-js@3.3.1` is out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,8 @@
   },
   "types": "dist/types/index.d.ts",
   "browser": {
+    "readable-stream": false,
+    "stream": false,
     "./src/utils/data.ts": "./src/utils/data.browser.ts",
     "./dist/cjs/utils/data.js": "./dist/cjs/utils/data.browser.js",
     "./dist/mjs/utils/data.js": "./dist/mjs/utils/data.browser.js",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -71,13 +73,15 @@
...
-    "readable-stream": "^3.6.0",
     "semver": "^7.3.5",
     "tar-js": "^0.3.0",
     "utf-8-validate": "^5.0.8",
     "web-streams-polyfill": "^3.1.0",
     "ws": "^8.5.0"
   },
+  "peerDependencies": {
+    "readable-stream": "^3.6.0"
+  },
```
vojtechsimetka, line 84: I think you need to add this otherwise project compilations will yell if there is no `readable-stream` installed
```json
"peerDependenciesMeta": {
  "readable-stream": {
    "optional": true
  }
}
```

AuHau: Hmm thanks for the info!

vojtechsimetka (approved):
LFTM

PR #562, opened 2022-02-15
fix: add js suffix in imports for esm support with babel

### LINKED ISSUE #555: Cannot bundle with webpack
> npm i  @ethersphere/bee-js webpack webpack-cli

> src/index.js
```js
import { Bee } from '@ethersphere/bee-js'
console.log(Bee)
```

And run `npx webpack`, it reports:

```
WARNING in ./src/index.js 2:12-15
export 'Bee' (imported as 'Bee') was not found in '@ethersphere/bee-js' (module has no exports)

ERROR in ./node_modules/.pnpm/@ethersphere+bee-js@3.3.0/node_modules/@ethersphere/bee-js/dist/mjs/index.js 1:0-28
Module not found: Error: Can't resolve './bee' in 'C:\Users\Jack\Workspace\bee\node_modules\.pnpm\@ethersphere+bee-js@3.3.0\node_modules\@ethersphere\bee-js\dist\mjs'
Did you mean 'bee.js'?
BREAKING CHANGE: The request './bee' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
 @ ./src/index.js 1:0-41 2:12-15
```

It looks like this package ships an invalid ESM bundle. (Node ESM requires a fully specified extension name).

AuHau:
Potentially superseeds the #561 with using Babel to add the `.js` suffix and hopefully still functional tests.

Closes #555

### DIFF
--- a/src/modules/debug/status.ts
+++ b/src/modules/debug/status.ts
@@ -2,7 +2,7 @@ import { http } from '../../utils/http'
 import type { Health, NodeInfo } from '../../types/debug'
 import { Ky } from '../../types'
 import { BeeVersions } from '../../types/debug'
-import getMajorSemver from 'semver/functions/major'
+import getMajorSemver from 'semver/functions/major.js'
 
 // Following lines bellow are automatically updated with GitHub Action when Bee version is updated
 // so if you are changing anything about them change the `update_bee` action accordingly!
--- a/.depcheckrc.json
+++ b/.depcheckrc.json
@@ -2,6 +2,7 @@
   "ignores": [
     "@babel/core",
     "@babel/plugin-proposal-class-properties",
+    "babel-plugin-add-import-extension",
     "@babel/plugin-transform-runtime",
     "@babel/preset-env",
     "@babel/preset-typescript",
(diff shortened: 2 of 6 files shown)


src/modules/debug/status.ts
```diff
@@ -2,7 +2,7 @@ import { http } from '../../utils/http'
 import type { Health, NodeInfo } from '../../types/debug'
 import { Ky } from '../../types'
 import { BeeVersions } from '../../types/debug'
-import getMajorSemver from 'semver/functions/major'
+import getMajorSemver from 'semver/functions/major.js'
```
vojtechsimetka, line 5: Just for me to understand. These library imports have to have `.js` appended because they are also ESM? So if some other dependency (e.g. `js-sha3`) gets updated and bundeld as ESM, we would also add the suffix there?

vojtechsimetka (approved):
LGTM, definitely like it better than specifying `.js` suffixes in the TS files :)

PR #570, opened 2022-02-18
ci: configs syncing

AuHau:


vojtechsimetka (approved):
The action LGTM. 

IMHO there is files in https://github.com/ethersphere/configs-js that should not be there as is are very specific to `bee-js` - `depcheckrc.json`. Please consider removing it.

PR #599, opened 2022-03-21
chore: update to the bee api SwarmCommon v2 and Swarm v3

vojtechsimetka:
**Important: we decided to not merge this until next API breaking change.** The reasoning is, that the current bee-js works both for Bee v1.4 and v1.5 and by merging this PR we would stop supporting bee v1.4.

The bee v1.5.0 replaces the null responses with empty arrays on successful requests that have no result. With bee-js we have been already returning empty arrays for such responses but now we no longer need to do this check.

https://github.com/ethersphere/bee/pull/2821

### DIFF
--- a/src/modules/debug/connectivity.ts
+++ b/src/modules/debug/connectivity.ts
@@ -19,7 +19,7 @@ export async function getPeers(ky: Ky): Promise<Peer[]> {
     responseType: 'json',
   })
 
-  return response.data.peers || []
+  return response.data.peers
 }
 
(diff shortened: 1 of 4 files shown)


vojtechsimetka:
@AuHau should we finally merge this? :)

AuHau (approved):

bee-worker:
Release [v5.0.0](https://github.com/ethersphere/bee-js/releases/tag/v5.0.0) addresses this.

PR #614, opened 2022-04-04
fix: remove browser blob polyfill for service workers support

### LINKED ISSUE #586: Bee-js cannot be executed in service worker script
When the library is loaded in a service worker script, the following error occurs:

`Uncaught ReferenceError: document is not defined`

The reason is that the library includes the `blob-polyfill` which tries to access the document object.

**Note:**
One temporary solution is to replace that polyfill with an empty object, in webpack config:

```javascript
plugins: [
  new NormalModuleReplacementPlugin(
      /blob-polyfill/,
      path.join(srcDir, 'polyfills', 'Blob.js')
    ),
]
```

src/polyfills/Blob.js
```javascript
module.export = {};
```

AuHau:
Closes #586

### DIFF
--- a/.depcheckrc.json
+++ b/.depcheckrc.json
@@ -17,6 +17,7 @@
     "babel-jest",
     "babel-loader",
     "bufferutil",
+    "cross-blob",
     "debug",
     "husky",
     "jest-puppeteer",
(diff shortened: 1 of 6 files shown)


vojtechsimetka (approved):
LGTM

bee-worker:
Release [@ethersphere/bee-js v3.3.4](https://github.com/ethersphere/bee-js/releases/tag/v3.3.4) addresses this.

PR #630, opened 2022-04-13
refactor: simplifying feeds component

AuHau:
This refactor mainly aims on simplifying the `feeds/index.ts` file and it does so with:

 - deduping generic constants using already existing ones in the `types/index.ts` file
 - creating `BytesReference` type that handles references in bytes
 - moving chunk specific functions into more appropriate places
 - Splitting identifier functionality into separate file

### DIFF
--- a/src/feed/index.ts
+++ b/src/feed/index.ts
@@ -1,38 +1,36 @@
 import { keccak256Hash } from '../utils/hash'
 import { serializeBytes } from '../chunk/serialize'
-import { Identifier, uploadSingleOwnerChunkData, makeSingleOwnerChunkFromData } from '../chunk/soc'
 import { FeedUpdateOptions, fetchLatestFeedUpdate, FetchFeedUpdateResponse } from '../modules/feed'
+import { makeSingleOwnerChunkFromData, uploadSingleOwnerChunkData } from '../chunk/soc'
 import {
-  REFERENCE_HEX_LENGTH,
-  Reference,
-  UploadOptions,
-  ENCRYPTED_REFERENCE_HEX_LENGTH,
-  ENCRYPTED_REFERENCE_BYTES_LENGTH,
-  REFERENCE_BYTES_LENGTH,
-  Signer,
-  FeedReader,
-  FeedWriter,
-  Topic,
   Address,
   BatchId,
+  BytesReference,
+  FEED_INDEX_HEX_LENGTH,
+  FeedReader,
+  FeedWriter,
   Ky,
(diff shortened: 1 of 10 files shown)


src/feed/index.ts
```diff
@@ -125,38 +68,37 @@ export async function updateFeed(
...
+  index: Index = 'latest',
 ): Promise<Reference> {
   const ownerHex = makeHexEthAddress(signer.address)
-  const nextIndex = await findNextIndex(ky, ownerHex, topic, options)
+  const nextIndex = index === 'latest' ? await findNextIndex(ky, ownerHex, topic, options) : index
 
-  return uploadFeedUpdate(ky, signer, topic, nextIndex, reference, postageBatchId, options)
-}
+  const identifier = makeFeedIdentifier(topic, nextIndex)
+  const at = options?.at ?? Date.now() / 1000.0
```
vojtechsimetka, line 80: interesting, I had to google this operator. Would you say, that there could just as well be `||` instead of `??` ?

src/utils/reference.ts
```diff
@@ -0,0 +1,41 @@
...
+      const hexReference = makeHexString(reference, ENCRYPTED_REFERENCE_HEX_LENGTH)
+
+      return hexToBytes<typeof ENCRYPTED_REFERENCE_BYTES_LENGTH>(hexReference)
+    }
+  } else if (reference instanceof Uint8Array) {
+    try {
+      return bytesAtOffset(reference, offset, ENCRYPTED_REFERENCE_BYTES_LENGTH)
+    } catch (e) {
+      return bytesAtOffset(reference, offset, REFERENCE_BYTES_LENGTH)
+    }
```
vojtechsimetka, line 38: I don't quite like using the try/catch mechanism here. Couldn't this be checked with `if` statement?

vojtechsimetka (approved):

src/feed/index.ts
AuHau: `??` is a better `||` because it correctly handles falsy values that are otherwise correct values. For this case for example if the user would pass `options.at = 0` then with `||` you would actually get the `Date.now() / 1000.0` instead of `0` ;-)

vojtechsimetka: Yeah actually i like it better. I did check the documentation and indeed could think of several places where such check would make sense. But I would argue here that the “at” being 0 here is an error either way 😅. Anyways please keep it as is :)

vojtechsimetka (approved):

PR #659, opened 2022-05-03
feat: ens support for download methods

### LINKED ISSUE #647: Support ENS on download functions
Currently, if a user tries to download resource under ENS domain with bee-js, the download function will complain as it expects only `64` or `128` hexstrings.

```
Error: value not valid hex string of length 128: swarm.eth
```

This blocks https://github.com/ethersphere/bee-dashboard/pull/351

AuHau:
Closes #647


src/utils/type.ts
```diff
@@ -100,6 +103,31 @@ export function assertReference(value: unknown): asserts value is Reference {
   }
 }
 
+export function assertReferenceOrEns(value: unknown): asserts value is ReferenceOrEns {
+  if (typeof value !== 'string') {
+    throw new TypeError('ReferenceOrEns has to be a string!')
+  }
+
+  if (value.endsWith('.eth')) {
```
AuHau, line 111: I am not sure if this is a good approach to assert ENS domain?

As for example wrote here https://docs.ens.domains/ens-deployments there are different types of TLDs supported and also as the user of Bee can configure his own ENS resolver there can be custom TLD as well.

Maybe regex like this `.+\.[a-zA-Z]+$` would be better?

test/unit/assertions.ts
```diff
@@ -225,6 +225,39 @@ export function testReferenceAssertions(executor: (input: unknown) => void): voi
...
+    // Length mismatch
+    await expect(() => executor('4fb5a872396d9693e5c9f9d7233cfa93f395c093371017ff44aa9ae6564cdd')).rejects.toThrow(
+      TypeError,
+    )
+
+    // ENS with invalid characters
+    await expect(() => executor('⶿.ß‍.eth')).rejects.toThrow(BeeArgumentError)
+
+    // ENS with invalid characters
+    await expect(() => executor('some space.eth')).rejects.toThrow(BeeArgumentError)
```
vojtechsimetka, line 257: Please could you test other invalid hostname like `-example.eth` or `example-.eth` and also check whether it supports subdomains e.g. `subdomain.example.eth` ? Does it expect only the hostname or can it deal with additional info like protocol as well? (it probably should not)

src/utils/type.ts
vojtechsimetka: That is great question. IMHO right now we only support ENS.

But I would consider using regex instead of function. The regex you would need is `/^((?:(?!-)[a-z0-9-]{1,63}(?<!-)\.)+eth)$/i`. This supports subdomains as well and we could potentially add other domains like `crypto` by simply modifying the regex with `/^((?:(?!-)[a-z0-9-]{1,63}(?<!-)\.)+(?:eth|crypto))$/i`

test/unit/assertions.ts
```diff
@@ -225,6 +225,39 @@ export function testReferenceAssertions(executor: (input: unknown) => void): voi
...
+      TypeError,
+    )
+
+    // Length mismatch
+    await expect(() => executor('4fb5a872396d9693e5c9f9d7233cfa93f395c093371017ff44aa9ae6564cdd')).rejects.toThrow(
+      TypeError,
+    )
+
+    // ENS with invalid characters
+    await expect(() => executor('⶿.ß‍.eth')).rejects.toThrow(BeeArgumentError)
```
vojtechsimetka, line 254: I think we should test these separately
```suggestion
    await expect(() => executor('⶿.eth')).rejects.toThrow(BeeArgumentError)
    await expect(() => executor('ß‍.eth')).rejects.toThrow(BeeArgumentError)
```

test/unit/assertions.ts
vojtechsimetka: Yeah if it can deal with someone just pasting `http://example.eth` or if this is error and it recognizes only `example.eth`. This should imho be also explicitly tested.

src/utils/type.ts
vojtechsimetka: so after the daily discussion, the regex is `/^((?:(?!-)[a-z0-9-]{1,63}(?<!-)\.)+(?:(?!-)[a-z0-9-]{2,63}(?<!-))$/i`

test/unit/assertions.ts
AuHau: Hmm I am not sure about this. I took it from the test coverage of the package, so I would just keep it like it: https://github.com/hexonet/idna-uts46/blob/master/test/IdnaTest.txt#L6088

vojtechsimetka: Yeah I was assuming this tests non-ASCII chars but it actually tests more than that...

src/utils/type.ts
AuHau: In the end, I restructured the assert function so the domain assertion is the last thing and it is let up to the library to verify it is valid name without enforcing any suffix.

AuHau:
Btw. the `idna-uts46-hx` package is quite a big one 😳 It adds 160kb minified. See our deps including this package [here](https://bundlephobia.com/scan-results?packages=@types/readable-stream@2.3.13,bufferutil@4.0.6,elliptic@6.5.4,fetch-blob@2.1.2,idna-uts46-hx@4.0.0,isomorphic-ws@4.0.1,js-sha3@0.8.0,ky@0.25.1,ky-universal@0.8.2,semver@7.3.5,tar-js@0.3.0,utf-8-validate@5.0.9,web-streams-polyfill@4.0.0-beta.1,ws@8.5.0&sortMode=size).

vojtechsimetka (approved):

src/utils/type.ts
```diff
@@ -100,6 +101,42 @@ export function assertReference(value: unknown): asserts value is Reference {
...
+   * -adg.ets - INVALID
+   * adg-.ets - INVALID
+   * as-a.com - VALID
+   * ethswarm.org - VALID
+   * http://asdf.asf - INVALID
+   * řš+ýí.šě+ř.čě - VALID
+   * tsg.asg?asg - INVALID
+   * tsg.asg:1599 - INVALID
+   */
+  const DOMAIN_REGEX = /^(?:[^-.\/?:\s][^.\/?:\s]{0,62}\.)+[^.\/?:\s]{2,63}$/
```
vojtechsimetka, line 130: Small improvement because what you have here actually matches `adg-.ets` as valid and also TLDs which start or end with `-`
```suggestion
  /**
   * a.asdf - VALID
   * test.eth - VALID
   * ADAM.ETH - VALID
   * ADAM UHLIR.ETH - INVALID
   * test.whatever.eth - VALID
   * -adg.ets - INVALID
   * adg-.ets - INVALID
   * as-a.com - VALID
   * ethswarm.org - VALID
   * http://asdf.asf - INVALID
   * řš+ýí.šě+ř.čě - VALID
   * tsg.asg?asg - INVALID
   * tsg.asg:1599 - INVALID
   * ethswarm.something- - INVALID
   * ethswarm.-something - INVALID
   * ethswarm.some-thing - VALID
   */
  const DOMAIN_REGEX = /^(?:(?!-)[^.\/?:\s]{1,63}(?<!-)\.)+(?!-)[^.\/?:\s]{2,63}(?<!-)$/
```

test/unit/assertions.ts
```diff
@@ -225,6 +225,38 @@ export function testReferenceAssertions(executor: (input: unknown) => void): voi
...
+    // Length mismatch
+    await expect(() => executor('4fb5a872396d9693e5c9f9d7233cfa93f395c093371017ff44aa9ae6564cdd')).rejects.toThrow(
+      TypeError,
+    )
+
+    // ENS with invalid characters
+    await expect(() => executor('')).rejects.toThrow(TypeError)
+    await expect(() => executor('some space.eth')).rejects.toThrow(TypeError)
+    await expect(() => executor('-example.eth')).rejects.toThrow(TypeError)
+    await expect(() => executor('http://example.eth')).rejects.toThrow(TypeError)
```
vojtechsimetka, line 256: maybe this should test the values from the comment above the `DOMAIN_REGEX` https://github.com/ethersphere/bee-js/pull/659/files#diff-a2b7fd28204be52f46a027c1f4d5d513964cb30439c9d216032a0ab8bfcb7e41R130

vojtechsimetka (approved):

bee-worker:
Release [v4.0.0](https://github.com/ethersphere/bee-js/releases/tag/v4.0.0) addresses this.

PR #665, opened 2022-05-13
chore: update to bee 1.6.0-6ceadd35

bee-worker:
Updated Bee version 1.6.0-6ceadd35

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,8 +134,8 @@
   "engines": {
     "node": ">=12.0.0",
     "npm": ">=6.0.0",
-    "beeApiVersion": "3.0.0",
-    "beeDebugApiVersion": "2.0.0",
-    "bee": "1.5.1-d0a77598"
+    "beeApiVersion": "3.0.1",
+    "beeDebugApiVersion": "2.0.1",
+    "bee": "1.6.0-6ceadd35"
   }
 }
--- a/src/modules/debug/status.ts
+++ b/src/modules/debug/status.ts
@@ -6,9 +6,9 @@ import getMajorSemver from 'semver/functions/major.js'
 
 // Following lines bellow are automatically updated with GitHub Action when Bee version is updated
 // so if you are changing anything about them change the `update_bee` action accordingly!
-export const SUPPORTED_BEE_VERSION_EXACT = '1.5.1-d0a77598'
-export const SUPPORTED_API_VERSION = '3.0.0'
-export const SUPPORTED_DEBUG_API_VERSION = '2.0.0'
+export const SUPPORTED_BEE_VERSION_EXACT = '1.6.0-6ceadd35'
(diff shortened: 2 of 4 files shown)


bee-worker:
## 🗒 Bee 1.6.0-6ceadd35 Release notes

This latest release is an accumulation of a lot of work done in the last couple of months by the team. We've improved and refactored substantial
amount of components and areas of the codebase as well as fixed a few critical bugs. After deploying this version your node will be forced into a
blockchain resync, so please allow it time to sync and in case you're updating several nodes at the same time - make sure to shelf the update in a way
that does not cause a lot of them to go offline at the same time due to the long restart time caused by chain syncing.


## Breaking changes

- Global pinning has been discontinued from bee. We're looking for ways to enable this sort of solution as an L2 solution on Swarm (#2906, #2869)
- Nodes will now try to connect to the mainnet by default. In order to connect to the testnet you must specify `--mainnet=false` (#2859)

## Deprecations


## Hardening

- Chain syncing errors will again result in 24h blocking (#2922)
- We've improved resilience on hibernation and network failure (#2887)
- Added a missing stream reset stream on add context header fail (#2885)

## New features
- Added a new db indices command to debug database indices (#2924)
- Added a wallet endpoint to display wallet information for various purposes (#2900)
- Exposed current block height from chain backend for the chainstate endpoint so that chain syncing status can be displayed (#2893)
- The node will now log its own overlay address on startup (#2930)
- Print required funds for gas and respect deployment gas price (#2861)
- The node has now an option to bootstrap the postage information partially from a snapshot stored on Swarm(!). This saves on the initial syncing time significantly, and reduces the amount of chain calls you would need to do (#2814, #2910)

## Bugfixes
- Added fallback content length check in middleware (#2907)
- Fixed a bug where unpinning of encrypted references resulted in an error (#2896)
- Fixed a panic for the stewardship endpoint in dev mode (#2889)
- CORS are now allowed in dev mode (#2866)
- Fixed a deadly deadlock in sharky that was caused when a context was cancelled on a Read call (#2932, #2948)
- Fixed a bug where an error on shutdown would cause issues when reloading stamp issuers on startup (#2914)
- Fixed a bug where the wrong amount was displayed when purchasing a postage stamp in dev mode (#2945)
- Removed the free slot optimization from sharky to circumvent shard leakage (#2953)

## Performance improvements
- The postage package has been rewritten and has significant simplifications that should result in a smoother and more correct operation (#2810, #2890, #2895)

For a full PR rundown please consult the `v1.6.0` [milestone](https://github.com/ethersphere/bee/milestone/15?closed=1).

AuHau (approved):

bee-worker:
Release [v4.0.0](https://github.com/ethersphere/bee-js/releases/tag/v4.0.0) addresses this.

PR #681, opened 2022-05-19
feat: cids support

### LINKED ISSUE #648: Support Swarm CID references
Swarm has already CID [library](https://github.com/ethersphere/swarm-cid-js/) and also CID based linking `{cid}.bzz.link`.
Would be awesome, if bee-js also supported this kind of reference.

AuHau:
Closes #648 

The Feed CID is not returned for the `Bee.createFeedManifest()` as that would be breaking change. It is planned for the next breaking release.

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -62,9 +62,13 @@ export type Reference = HexString<typeof REFERENCE_HEX_LENGTH> | HexString<typeo
 
 /**
  * Type that represents either Swarm's reference in hex string or ESN domain (something.eth).
- * This is the type used on all the download functions.
  */
 export type ReferenceOrEns = Reference | string
+
+/**
+ * Type that represents either Swarm's reference in hex string, ESN domain (something.eth) or CID using one of the Swarm's codecs.
+ */
+export type ReferenceCidOrEns = ReferenceOrEns | string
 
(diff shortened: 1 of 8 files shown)


src/types/index.ts
```diff
@@ -62,10 +62,14 @@ export type Reference = HexString<typeof REFERENCE_HEX_LENGTH> | HexString<typeo
...
 /**
  * Type that represents either Swarm's reference in hex string or ESN domain (something.eth).
- * This is the type used on all the download functions.
  */
 export type ReferenceOrEns = Reference | string
 
+/**
+ * Type that represents either Swarm's reference in hex string, ESN domain (something.eth) or CID using one of the Swarm's codecs.
+ */
+export type ReferenceCidOrEns = ReferenceOrEns | string
```
vojtechsimetka, line 71: I assume this is just to make it a bit more explicit in the code. Because the `ReferenceCidOrEns` is the same type as `ReferenceOrEns`

vojtechsimetka (approved):

src/types/index.ts
AuHau: Yup, I was thinking about using the `BrandedType` but then decided to go only with this.

bee-worker:
Release [v4.1.0](https://github.com/ethersphere/bee-js/releases/tag/v4.1.0) addresses this.

PR #685, opened 2022-05-22
ci: node 18 testing

AuHau:


.github/workflows/tests.yaml
```diff
@@ -54,7 +54,13 @@ jobs:
...
       # Start Bee Factory environment
       - name: Start Bee Factory environment
-        run: npm run bee -- --detach
+        if:  matrix.node-version != '18.x'
+        run: npm run bee -- --verbose --detach
+
+      # Start Bee Factory environment for Node 18
+      - name: Start Bee Factory environment for Node 18
+        if:  matrix.node-version == '18.x'
+        run: node --no-experimental-fetch ./node_modules/.bin/bee-factory start --verbose --detach
```
vojtechsimetka, line 63: Isn't there environment variable rather than flag we can set instead of starting the bee factory this way?

vojtechsimetka (approved):

.github/workflows/tests.yaml
AuHau: I thought not, but actually yes.

bee-worker:
Release [v4.1.0](https://github.com/ethersphere/bee-js/releases/tag/v4.1.0) addresses this.

PR #696, opened 2022-05-26
fix: move sleep to have all code in src

AuHau:
So... 😅

My IDE was trigger-happy and I haven't noticed it. When implementing the `waitForUsable` feature, I have copy&pasted Aron's solution, which used the `sleep` function and IDE noticed that we have this function in `test/utils.ts` file so it just imported that and I have not noticed that.

Thanks to this the `tsc` compilation changed the structure of the compiled files as it has to now import something from `./test` folder and so it changed the `./dist/cjs/` folder structure into containing subfolders `src` and `test` as this:

```
cjs
├── package.json
├── src
│   ├── bee-debug.js
│   ├── bee.js
│   ├── chunk
│   ├── feed
│   ├── index.js
│   ├── modules
│   ├── types
│   └── utils
└── test
    └── utils.js
```

This, unfortunately, breaks the package as the entry file `./dist/cjs/index.js` is not in the expected place.

### DIFF
--- a/src/bee-debug.ts
+++ b/src/bee-debug.ts
@@ -62,7 +62,7 @@ import * as tag from './modules/debug/tag'
 import * as stamps from './modules/debug/stamps'
 import type { Options as KyOptions } from 'ky-universal'
 import { makeDefaultKy, wrapRequestClosure, wrapResponseClosure } from './utils/http'
-import { sleep } from '../test/utils'
+import { sleep } from './utils/sleep'
 
 export class BeeDebug {
   /**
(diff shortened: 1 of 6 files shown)


vojtechsimetka (approved):

bee-worker:
Release [v4.1.1](https://github.com/ethersphere/bee-js/releases/tag/v4.1.1) addresses this.

agazso:
Maybe we can add a test for this in the release pipeline so we don't accidentally break this in the future?

Issue #730, opened 2022-07-07
`addCidConversionFunction` seems to break encrypted uploads


Cafe137:
```
npm i @ethersphere/bee-js
```

```
const STAMP = '...'
const bee = new Bee('http://localhost:1633')
bee.uploadFile(STAMP, 'util.js', 'util.js', { encrypt: true })
```

```
TypeError: Reference does not have expected length 64 characters. Encrypted references are not supported.
```

AuHau:
Hmm yeah, Swarm's CIDs are not supported for encrypted references as they are too long to encode. I guess you are automatically calling the `.cid` property on the uploaded result? If so then you have to remove that on encrypted uploads...

Cafe137:
It seems to run automatically:

```
TypeError: Reference does not have expected length 64 characters. Encrypted references are not supported.
    at assertReference (/Users/aron/Code/swarm-cli/dist/index.js:16490:11)
    at _encodeReference (/Users/aron/Code/swarm-cli/dist/index.js:16415:3)
    at encodeReference (/Users/aron/Code/swarm-cli/dist/index.js:16325:14)
    at Object.get cid [as cid] (/Users/aron/Code/swarm-cli/dist/index.js:108736:46)
    at Function.assign (<anonymous>)
    at addCidConversionFunction (/Users/aron/Code/swarm-cli/dist/index.js:108734:17)
    at Bee.<anonymous> (/Users/aron/Code/swarm-cli/dist/index.js:101538:52)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/aron/Code/swarm-cli/dist/index.js:101250:24)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

bee-worker:
Release [v5.0.0](https://github.com/ethersphere/bee-js/releases/tag/v5.0.0) addresses this.

PR #735, opened 2022-07-19
fix: do not call cid automatically on encrypted upload results

Cafe137:


Cafe137:
Fixes #730

AuHau (approved):
That is sweet! Thanks a lot for the fix. In hindsight I guess the getter was not the best design choice I made...

agazso:
> That is sweet! Thanks a lot for the fix. In hindsight I guess the getter was not the best design choice I made...

Can we maybe discuss the alternatives and have a proposal for a better design than to get stuck with something you describe as `not the best design choice`? 😄

Cafe137:
However this does not comply with the tests yet, so I need to come up with a better solution.

AuHau:
Superceeded by #740 so closing

PR #772, opened 2022-09-29
feat!: esm only build

AuHau:
This PR implements ESM-only support, this on its own is breaking change, but together with it it brings several other changes:

 - the browser bundle is not supposed to be imported like for example `await import('https://unpkg.com/@ethersphere/bee-js/dist/index.browser.js')`
 - the `Data` interface used to be extended `Uint8Array` now in order to get `Uint8Array` you need to use `Data.bytes()` function

Great write-up about possible issues, how-to guides, motivation etc. done by 
Sindre Sorhus: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

One issue of migrating to ESM is about imports, which now has to have the full file name **including the suffix**. A bit non-intuitive thing is that even though we work in the TypeScript realm, the import must have `.js` suffix. TypeScript can work with it, but unfortunately does not provide a way to transpile `.ts` suffix to `.js` or add it based on configuration. There are three ways how to solve this:

 - for new projects, just have `import`s with `.js` suffix (we should figure out ESLint rule for that)
 - for existing Node backend projects that uses something like `ts-node` then using [`--experimental-specifier-resolution=node`](https://nodejs.org/api/esm.html#customizing-esm-specifier-resolution-algorithm) Node flag, lifts this requirement, although it is good to be cautious about this as this is marked as an experimental flag
 - for libraries (also can be applied for backends) use the `babel-plugin-add-import-extension` on top of TypeScript transpiled sources. For example see [here](https://github.com/ethersphere/bee-js/blob/4ef319822393f19b796d66cedfaa328c7331e6cd/package.json#L49)


.eslintrc.cjs
```diff
@@ -89,6 +89,7 @@ module.exports = {
     ],
     'require-await': 'off',
     '@typescript-eslint/promise-function-async': 'error',
+    // '@typescript-eslint/explicit-function-return-type': 'error',
```
Cafe137, line 92: Do we need this here?

src/types/index.ts
```diff
@@ -321,7 +321,12 @@ export interface Pin {
  *
  * Concretely: text(), hex(), json()
```
Cafe137, line 322: Add `array()` here?

src/types/index.ts
```diff
@@ -321,7 +321,12 @@ export interface Pin {
...
  */
-export interface Data extends Uint8Array {
+export interface Data {
+  /**
+   * The raw binary data in form of Uint8Array
+   */
+  array(): Uint8Array
```
Cafe137, line 328: I was confused about the name `array` before I have seen this method, maybe `bytes` would be a more descriptive name, albeit the JSDoc makes it clear quickly.

AuHau: Yeah `bytes()` is a better name indeed. I will change it.

Cafe137 (approved):

PR #774, opened 2022-10-04
chore: update ky

AuHau:


src/utils/http.ts
```diff
@@ -18,7 +20,7 @@ interface UndiciError {
 }
 
 interface KyResponse<T> extends Response {
-  data: T
```
AuHau, line 21: `ky` has `data` property protected using exposing only a getter and not a setter, so we have to change the property that we use to parse the response data.

package.json
```diff
@@ -88,8 +88,9 @@
     "@commitlint/config-conventional": "^17.1.0",
     "@ethersphere/bee-factory": "^0.4.1",
     "@fluffy-spoon/substitute": "^1.208.0",
-    "@jest/test-sequencer": "^29.1.0",
-    "@jest/globals": "^29.1.1",
+    "@jest/globals": "^29.1.2",
+    "@jest/test-sequencer": "^29.1.2",
+    "@naholyr/cross-env": "^1.0.0",
```
AuHau, line 93: I had to use a fork of `cross-env` that allows setting defaults values for the env. variables, in order to allow overriding them (for Node18 where we need to disable the native `fetch`).

Cafe137 (approved):

PR #782, opened 2022-10-21
fix: node18 native blob support

AuHau:
Node 18 ships with native Fetch that also exports its own Blob implementation. If the native `fetch` is used then it complains if it receives the `fetch-blob` polyfil instead of its own Blob implementation. This PR fixes this issue.


Cafe137 (approved):

AuHau:
So I have been testing it a bit more and I am not sure what was the original issue about, as I can't replicate the premise that native fetch does not accept the `blob-polyfill` anymore 😳 So closing this PR. Will see if the issue will re-appear or not.

PR #803, opened 2022-12-21
fix: backward compatibility for wallet ep

AuHau:


AuHau:
Tests won't be passing because of the Staking break, that is not fixed yet.

Cafe137 (approved):

Issue #837, opened 2023-02-20
"Segmentation fault (core dumped)" on soc uploading using Typescript/Jest


IgorShadurin:
Encountered an issue while using bee-js 5.2.0, specifically when uploading on SOC in Typescript + Jest project. The error received is a `Segmentation fault (core dumped)` on Ubuntu/MacOS.
There is test project: https://github.com/IgorShadurin/bee-js-ts
Node: v16.17.0

Node + js, Node + ts and web are working well.

```
# install dependencies
npm ci

# run test
npm run test

# run test with experiment flag
npm run test-e
```

PR #842, opened 2023-03-07
feat: change http library

Cafe137:
Previous import workaround causes segfault, which tracked down is an ever-open engine bug https://github.com/ethersphere/bee-js/issues/837

- [ ] Re-add before/after hooks
- [ ] Re-add retry / timeout
- [ ] Re-add custom errors
- [ ] Cleanup public interface of request options
- [ ] Cleanup utils


bee-worker:
Release [v6.0.0](https://github.com/ethersphere/bee-js/releases/tag/v6.0.0) addresses this.

Issue #360, opened 2021-06-24
Support streaming upload of Directories


AuHau:
Currently, it is possible to upload in streaming way file with the `uploadFiles()` method when `Readable` is passed. This is possible only in NodeJs at the moment as browsers currently does not support streaming requests. This will hopefully change soon as for example, Chrome has already experimental support for this (https://web.dev/fetch-upload-streaming/).

But... The current implementation of upload of directories relies on the creation of Tar before the upload. This is done all in the memory, so if for example uploading a large directory to Swarm, then everything needs to be loaded into memory, which might lead to Out of Memory errors for Nodejs. It is possible to create Tar in streaming way with [`tar-stream`](https://www.npmjs.com/package/tar-stream). We should revisit this to add support for the streaming of directories.

PR #895, opened 2023-12-15
feat: tar-stream library instead of tar-js (#360)

tomicvladan:
Using the `tar-stream` library which creates tar as a stream.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "isomorphic-ws": "^4.0.1",
         "js-sha3": "^0.8.0",
         "semver": "^7.3.5",
-        "tar-js": "^0.3.0",
+        "tar-stream": "^3.1.6",
         "web-streams-polyfill": "^4.0.0-beta.3",
         "ws": "^8.7.0"
       },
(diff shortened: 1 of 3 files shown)


agazso:
Does this work in a browser?

tomicvladan:
No, this is just a fix for nodeJS to overcome size limitation.

AuHau:
You can make work `tar-stream` in a browser, but the NodeJs APIs (`Buffer` etc) need to by polyfilled. I have done it here: https://github.com/rsksmart/rif-storage-js/blob/master/src/swarm-mini/index.ts

But not sure if it is good idea. It is quite painful.

AuHau:
Oh right, exactly for this reason I pushed actually to implement streaming support, but it got never merged :-(

There is draft PR for that here: https://github.com/ethersphere/bee-js/pull/372

tomicvladan:
Using a polyfill for Node structures could be a solution, but I think it's more efficient to use browser streams:

https://developer.mozilla.org/en-US/docs/Web/API/Streams_API

It can work with Axios as well.

agazso:
From previous experience the Streams API have several issues regarding compatibility, ease of use and maintenance. Also using `tar` can be considered as a workaround from the Swarm API perspective to allow working with multiple folders and files.

Personally I think it would be more interesting and future-proof to use the Swarm chunk API as an alternative with the help of external projects (bmt-js, mantaray-js) as it matches better the semantics of the problem space and thus could open up new directions, provide better control and create optimization opportunities. But that solution may be a separate library in itself and thus outside of the scope of bee-js.

PR #906, opened 2024-02-27
fix: do not dispose arraybuffer in browser

Cafe137:


bee-worker:
Release [v6.7.4](https://github.com/ethersphere/bee-js/releases/tag/v6.7.4) addresses this.

PR #935, opened 2024-07-17
build: disable legacy peer deps

### LINKED ISSUE #693: On `web-streams-polyfill@4.0.0` release remove legacy peer deps
With the Node 18 update that automatically comes with `npm@8.9.0` this creates a problem for bee-js as in `npm@8.6.0` was released fix that performs some new peer dependencies validation (see https://github.com/npm/cli/issues/4664) which unfortunately affects us as we are depending on `web-streams-polyfill@4.0.0-beta.1`, but the `ky-universal` has peer dependency definition `>=2.0.0` which unfortunately does not incorporate pre-releases, so we are getting peer conflicts upon `npm install` or `npm ci`. The temporary solution is that I have enabled legacy peer dependency resolution and once the release `web-streams-polyfill@4.0.0` is out we should revert that and use the more strict current implementation.

Cafe137:
Closes #693

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-legacy-peer-deps=true


bee-worker:
Release [v7.1.1](https://github.com/ethersphere/bee-js/releases/tag/v7.1.1) addresses this.

PR #949, opened 2024-09-16
fix: handle bad request in feed index lookup

Cafe137:


bee-worker:
Release [v8.0.2](https://github.com/ethersphere/bee-js/releases/tag/v8.0.2) addresses this.

PR #982, opened 2025-03-10
fix: fail Reference.isValid for odd-length hex strings

Cafe137:


bee-worker:
Release [v9.0.1](https://github.com/ethersphere/bee-js/releases/tag/v9.0.1) addresses this.

PR #1039, opened 2025-06-23
fix: fix property merging when request body is a large buffer

Cafe137:


bee-worker:
Release [v9.4.1](https://github.com/ethersphere/bee-js/releases/tag/v9.4.1) addresses this.

PR #1187, opened 2026-06-11
chore: remove axios

slapec93:


src/bee.ts
```diff
@@ -506,7 +506,7 @@ export class Bee {
       const fileData = await fileArrayBuffer(data)
       const fileName = name ?? data.name
       const contentType = data.type
-      const fileOptions = { contentType, ...options }
+      const fileOptions = { ...options, contentType }
```
Cafe137, line 509: Why is this change needed? Just curious.

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@ethersphere/bee-js",
-  "version": "12.2.2",
+  "name": "@upcoming/bee-js",
```
Cafe137: This needs a revert

Cafe137 (changes requested):

src/bee.ts
slapec93: seems like Claude's sidequest:
> The old order let options override contentType, meaning if a caller explicitly passed a contentType in options, it would silently win over the file's own data.type. The new order makes the file's actual MIME type take precedence, which is the correct behavior — when you have a File object, its .type should not be overridable by the caller.

I'm not sure if it breaks anything, should I revert this change?

Cafe137: Ok, we can keep it, thanks

Cafe137:
@slapec93 I see that `throwIfAborted` has been removed from multiple places. Does that mean it works natively with `fetch` now, and glue code is no longer needed?

slapec93:
> I see that throwIfAborted has been removed from multiple places. Does that mean it works natively with fetch now, and glue code is no longer needed?
@Cafe137 yes, allegedly fetch handles this natively, but we have tests in abort.spec.ts, that proves this

Cafe137 (approved):

PR #1202, opened 2026-07-06
fix: mantaray marshal and unmarshal logic

slapec93:


Cafe137 (approved):
LGTM, but we will need more data-driven testing before we can release this

PR #1236, opened 2026-07-21
refactor!: use swarm core

slapec93:
### Before TODO:

- [ ] Publish swarm-core as @ethersphere/swarm-core in npm.js and use that instead of `@upcoming`


src/chunk/soc.ts
```diff
@@ -81,37 +34,13 @@ export function unmarshalSingleOwnerChunk(
...
-  return {
-    data,
-    identifier,
-    signature,
-    span,
-    payload,
-    address: socAddress,
-    owner: ownerAddress,
+  try {
+    return coreUnmarshalSingleOwnerChunk(data instanceof Bytes ? data.toUint8Array() : data, address)
```
Cafe137: `Bytes` comes from core, and `coreUnmarshalSingleOwnerChunk` is a core function too, so this logic should be in the core, being able to juggle with its own types.

src/manifest/manifest.ts
```diff
@@ -1,22 +1,64 @@
-import { Binary, MerkleTree, Optional, Uint8ArrayReader } from 'cafe-utility'
+import { Optional } from 'cafe-utility'
```
Cafe137, line 1: Manifest should also be in core. Bee-JS should only provide a thin layer on how to serialize, deserialize it recursively (so a Bee class, and the HTTP options params)

src/chunk/cac.ts
```diff
@@ -45,39 +40,21 @@ export interface Chunk {
...
+function withToSingleOwnerChunk(core: { data: Uint8Array; span: Span; payload: Bytes; address: Reference }): Chunk {
+  return {
+    data: core.data,
+    span: core.span,
+    payload: core.payload,
+    address: core.address,
+    toSingleOwnerChunk: (identifier, signer) =>
+      makeSingleOwnerChunk(core.address, core.span, core.payload, identifier, signer),
+  }
+}
```
Cafe137: Would it be possible to add this method to CAC in the core, too? Then this function, and some others can be removed from Bee-JS and stay fully supplied from core.

src/chunk/soc.ts
```diff
@@ -1,71 +1,24 @@
...
+  uint256ToNumber,
+  unmarshalSingleOwnerChunk as coreUnmarshalSingleOwnerChunk,
+} from 'swarm-core'
+
+export { makeSOCAddress }
+export type { SingleOwnerChunk }
+
+function privateKeyToBigInt(signer: PrivateKey): bigint {
+  return uint256ToNumber(signer.toUint8Array(), 'BE')
 }
```
Cafe137: Can be a method on `PrivateKey`

Cafe137 (approved):

PR #1239, opened 2026-07-28
fix: streamable data upload and preserve http request data

bosi95:
Extend function arguments for data.upload to enable streaming upload via the /bytes api endpoint.
Fix the missing request body for http calls that gets overwritten by Objects.deepMerge3.


Cafe137:
Handled in https://github.com/ethersphere/bee-js/pull/1249 due to permission issues

PR #1254, opened 2026-08-28
feat: rolling feed

slapec93:


src/modules/rolling-feed.ts
Cafe137, line 1: Filename should be rolling-feed.ts

README.md
```diff
@@ -332,6 +334,28 @@ const bee = new Bee('http://localhost:1633')
...
+const signer = new PrivateKey('...')
+const periodLength = 600 // 10 minutes
+
+const writer = bee.rollingFeed.makeWriter(topic, signer, periodLength)
+await writer.uploadPayload(batchId, 'Hello, World!')
+
+const reader = bee.rollingFeed.makeReader(topic, signer.publicKey().address(), periodLength)
+const result = await reader.downloadPayload()
+console.log(result.payload.toUtf8()) // prints 'Hello, World!'
+```
```
Cafe137, line 391: It would be better if we made it more obvious that a rolling feed writer should run on a timer / interval. Maybe split this sample to a writer and a reader part, and use `setInterval` in the writer to show a proper setup.

Cafe137 (changes requested):
Almost perfect, just a filename convention and a doc request

Cafe137 (approved):
LGTM

